PR #8, opened 2024-09-06
Revert "features/eu-debug: Initial patches for eu-debug support for L…

lucasdemarchi:
…NL/BMG"

This reverts commit c87b8510d26b7c74a2882ab035ba12e4545b48c7.

Do not mix for now the additional non-upstream features with backports. Particularly bad because with these changes CI is blocked.


ayazahma:
Instead of reverting, the script has been changed to build the base kernel so this pr is no longer needed. so closing the pr

PR #12, opened 2024-09-09
features/eu-debug: Remove kunit changes

kolanupa:
Adding the missing kunit related patches without which build is getting
fail on kunit config enablement.


ayazahma (changes requested):
pls update script to have eu debug tree  by default also update the Readme for script changees

ayazahma (approved):

lucasdemarchi:
there are regressions showing up in CI when you take the features in. Also, why would it be used by default? Hopefully we are not trying to send to Ubuntu by mistake on the generic kernel.

kolanupa:
> there are regressions showing up in CI when you take the features in. Also, why would it be used by default? Hopefully we are not trying to send to Ubuntu by mistake on the generic kernel.

Core BAT results are good,
Regression seen only on display tests especially for kms_psr* tests. Looks like all tests got passed but seen dmesg warn/crash in the logs. Hence the percentage got dropped. We will ask display team to have a look.

ayazahma:
we have clearly informed Ubuntu that Patches in base folder need to taken in generic-kernel and patches in features folder are for linux-intel kernel

PR #13, opened 2024-09-17
Rebase to 6.11 kernel

kolanupa:


lucasdemarchi (approved):
It matches my rebase too.  Since we didn't drop or added any patch on purpose, not much reason to send this elsewhere as the rebase can be handled downstream

PR #14, opened 2024-09-19
Compat with git-pile and quilt

lucasdemarchi:
Few changes to the format maintained here so it's compatible with git-pile and quilt.


kolanupa (approved):
LGTM

PR #41, opened 2025-01-13
features/eu-debug: Removes display patch which is missing from series

kolanupa:
Removing the display patch which is not included in series file.


mczaplin:
BAT testing went through, but there seems be some sort of issue when it comes to posting it automatically. I'm looking into it, and in the meantime the results can be found here: https://intel-gfx-ci.01.org/tree/xe-kmd-backports/xkb-pr-40-53ada6682063f7db2eb2984080d0c9d238754a22/index.html?

ayazahma (approved):

PR #43, opened 2025-01-24
Read the xkb tag and add it to generated tree

kolanupa:
Adding the xkb tag to generated kernel tree by reading it from the
xekmd-backports tree to identify from which HEAD it has generated.

### DIFF
--- a/backport.sh
+++ b/backport.sh
@@ -19,6 +19,8 @@ OPTS=$(getopt -a --options $SHORT --longoptions $LONG -- "$@")
 
 flavor="features"
 is_existing_tree=0
+xkb_tag=""
+
 usage () {
           echo ""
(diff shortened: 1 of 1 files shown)


kolanupa:
Tag format: xkb-<xekmd-backports tag>
    Ex: xkb-v6.11.11-241230.1
Tag format when there is no tag at HEAD: xkb-<tag>-<no.of commits>-g<commit sha of HEAD>
     Ex: xkb-v6.11.11-241230.1-2-gd86425a

ayazahma (changes requested):
We should not override kernel tree , we should have a option to override , otherwise just create a new branch 

also change the tag from xkb to xe

ayazahma (approved):

PR #48, opened 2025-03-10
features/eu-debug: To resolve hang type behavior

kolanupa:
Hang type behavior is seen instead of kernel bp. These patches are
needed to resolve it.


ayazahma:
Approved with condition of test passing or failure justified

Dominisg:
Afaik, interrupt-other sporadically fails on all our code bases.

kolanupa:
Looks like it is the random failure.
I have verified the BAT for LNL and looks no new regression found with this build.
Execution link: https://gtax-ril-fm.intel.com/#/jobs/3779877

PR #50, opened 2025-03-12
Update bat_trigger.yml

mczaplin:
Required fix for reporting bat testing comparisons


mczaplin:
@ayazahma please take a look so we can unblock reporting results

ayazahma (approved):

PR #75, opened 2025-06-26
patches/base: Fix for xe_force_gt_reset failure

pravalikagurram1:
This patch enables triggering a forced GT reset via a write operation.
Previously, GT resets were triggered by reading
the dedicated debugfs files "force_reset" and "force_reset_sync".
With this patch, a custom write handler is provided,
allowing GT resets to be triggered from user space using a write operation as well.


pravalikagurram1:
Closing  this PR because we are reverting the respective change in IGT

PR #79, opened 2025-07-10
features/eu-debug: Add DRM-managed cleanup support to avoid early release issues

kbommu:
The cleanup logic was previously called unconditionally during device teardown.
However, in some probe failure cases, this could happen before initialization
was completed, leading to failure.
To avoid this, cleanup is now registered as a DRM-managed action, only if
initialization succeeds. This ensures proper cleanup ordering and avoids
premature access to uninitialized structures.
Sysfs setup and teardown are updated similarly to follow the same safe and
managed pattern.


backport/patches/base/0001-drm-xe-uapi-Use-hint-for-guc-to-set-GT-frequency.patch
```diff
@@ -0,0 +1,302 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tejas Upadhyay <tejas.upadhyay@intel.com>
+Date: Fri, 28 Feb 2025 12:32:24 +0530
+Subject: drm/xe/uapi: Use hint for guc to set GT frequency
+
+Allow user to provide a low latency hint. When set, KMD sends a hint
+to GuC which results in special handling for that process. SLPC will
+ramp the GT frequency aggressively every time it switches to this
+process.
```
pravalikagurram1, line 9: What is this?
and commit description is not describing your change
Fixes: Bug#5137 - igt@xe_create@valid-flag fails due to missing bit in config->info[1]

pravalikagurram1: If it is open source bug you should mension like this 
please refer the previous commits

    Closes: https://gitlab.freedesktop.org/drm/xe/kernel/-/issues/5310

pravalikagurram1 (approved):

PR #246, opened 2026-01-08
drm/xe: Backport to fix potential use-after-scope bug in VF migration

pravalikagurram1:
Backport upstream fix to prevent use-after-scope bug where the loop
variable 'job' is referenced after the loop has exited. This can lead
to undefined behavior and potential crashes during VF (Virtual Function)
migration.

The issue occurs in the re-emission logic during VF migration where
the code attempts to access the loop iterator variable after the loop
completes. The fix saves the job from the last iteration of the loop
for use after the loop exits.

This backport resolves a potential stability issue in SR-IOV VF
migration scenarios.

Fixes: 29dbfa0a7 ("backport/sriov: Start re-emission from first unsignaled job during VF migration")


smuqthya:
For 6.14 we can proceed in sriov section as base if part of sriov. but this should have been in base

smuqthya (approved):

PR #250, opened 2026-01-21
drm/xe/pf: Handle VF BAR resize without PCI VF REBAR helpers

mbernatowicz:
Implement VF BAR resizing via the VF Resizable BAR capability directly
in PCI config space for kernels that lack PCI VF REBAR helper APIs.

### DIFF
--- a/src/drivers/gpu/drm/xe/xe_pci_sriov.c
+++ b/src/drivers/gpu/drm/xe/xe_pci_sriov.c
@@ -21,6 +21,44 @@
 #include "xe_sriov_pf_helpers.h"
 #include "xe_sriov_printk.h"
 
+/*
+ * NOTE: Backport-only.
+ *
+ * We need to update the SR-IOV VF BAR size cache (pdev->sriov->barsz[])
+ * when programming VF Resizable BAR (VF REBAR) on kernels that do not
+ * provide PCI VF REBAR helper APIs.
+ *
+ * The real 'struct pci_sriov' is PCI-core internal (drivers/pci/pci.h).
+ * For OOT/backport builds, we provide a local copy of the v6.6 layout
+ * to access barsz[] only. Keep this in sync with the target baseline.
+ */
+#ifndef HAVE_STRUCT_PCI_SRIOV
+struct pci_sriov {
+	int		pos;			/* Capability position */
+	int		nres;			/* Number of resources */
+	u32		cap;			/* SR-IOV Capabilities */
+	u16		ctrl;			/* SR-IOV Control */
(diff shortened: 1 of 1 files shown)


src/drivers/gpu/drm/xe/xe_pci_sriov.c
```diff
@@ -133,16 +171,73 @@ static void pf_engine_activity_stats(struct xe_device *xe, unsigned int num_vfs,
...
+{
+	return 1ULL << (size + ilog2((resource_size_t)SZ_1M));
+}
+#endif
+
 static int resize_vf_vram_bar(struct xe_device *xe, int num_vfs)
 {
 	struct pci_dev *pdev = to_pci_dev(xe->drm.dev);
-	u32 sizes;
+	resource_size_t size, total;
```
smuqthya, line 188: Is there a way to enable this code only when needed automatically. if not we should have a autotools enabled to verify at build time

src/drivers/gpu/drm/xe/xe_pci_sriov.c
```diff
@@ -294,6 +307,7 @@ static int pf_enable_vfs(struct xe_device *xe, int num_vfs)
 	return num_vfs;
 
 failed:
+	restore_vf_vram_bar_size(xe);
```
smuqthya, line 310: Does restoring required only when we resized it in the first place? if yes, then we should add it as part of control.

src/drivers/gpu/drm/xe/xe_pci_sriov.c
mbernatowicz: What u mean by that ? It's a version for 6.6 where we do not have PCI VF rebar helpers. Can u describe the process more ?

smuqthya: We may use same DKMS for version 6.6 till latest version of linux kernel and some time we may have to build against OS which may have backported the supported API's . Inorder to keep dkms unified, we are controlling the code based on the need of particular version. For example 6.6 may need this code, but if we are using Ubuntu Kernel with 6.17 and they backported the code.. we may not enable the code.

mbernatowicz: The initial ask was to provide alternate solution for 6.6 which lacks PCI VF rebar helpers and to put it in releases/main branch of backport repo, and it's the purpose of this patch.

smuqthya:
This need to be pushed to oot-backport/main . So cant merge in releases.

PR #266, opened 2026-02-06
gitignore: Exclude autotools and build-generated files

knayana-gse:
Ignore autotools artifacts and kernel configuration files that are
generated during the gentree and build process to prevent them from
being included in pull requests to releases/main.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,14 @@
+Makefile.in
+aclocal.m4
+autom4te.cache/
+configure
+compile
+install-sh
+missing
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,14 @@
+Makefile.in
```
smuqthya, line 1: Want to discuss more on this

knayana-gse: as discussed .gitignore is added in releases/main and it is part of PR #275.

PR #281, opened 2026-03-04
drm/xe: Backport upstream fixes set 3

kbommu:
Backport multiple bug fixes from upstream.

v2: Including the VF MCR-table initialization patch('drm/xe/vf: Allow VF
    to initialize MCR tables') is required together with:
    `drm/xe/wa: Steer RMW of MCR registers while building default LRC`,
    otherwise VF can hit MCR steering-table warnings during default LRC setup.

Signed-off-by: Bommu Krishnaiah <bommu.krishnaiah@intel.com>

### DIFF
--- a/series
+++ b/series
@@ -308,6 +308,14 @@ backport/patches/fixes/base/0001-drm-xe-nvm-Manage-nvm-aux-cleanup-with-devres.p
 backport/patches/fixes/base/0001-drm-xe-guc-Fix-CFI-violation-in-debugfs-access.patch
 backport/patches/fixes/sriov/0001-drm-xe-pf-Fix-sysfs-initialization.patch
 backport/patches/fixes/base/0001-drm-xe-bo-Redirect-faults-to-dummy-page-for-wedged-d.patch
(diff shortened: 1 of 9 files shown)


series
```diff
@@ -20,6 +20,7 @@ backport/patches/base/0001-drm-xe-dma-buf-Allow-pinning-of-p2p-dma-buf.patch
 backport/patches/base/0001-drm-xe-Drop-preempt-fences-when-destroying-imported-.patch
 backport/patches/base/0001-drm-xe-migrate-fix-batch-buffer-sizing.patch
 backport/patches/base/0001-drm-xe-make-xe_gt_idle_disable_c6-handle-the-forcewa.patch
+backport/patches/base/0001-drm-xe-vf-Allow-VF-to-initialize-MCR-tables.patch
```
pravalikagurram1: why does this is part of base folder?
we should not include anything in base folder correct?
and please add commit message which can speak about the process/change

kbommu: it is not fix patch, hence I included in base

pravalikagurram1: why fix patch is included in base?

kbommu: updated

pravalikagurram1 (approved):

PR #287, opened 2026-03-12
gitignore: Update to remove automacro.h file

knayana-gse:
Update gitignore file to exclude automacro.h file.

### DIFF
--- a/backport/.gitignore
+++ b/backport/.gitignore
@@ -37,3 +37,4 @@ Makefile.in
 config.log
 config.status
 src/backport-include/backport/backport_auto.h*
+src/backport-include/backport/automacro.h*


backport/.gitignore
```diff
@@ -37,3 +37,4 @@ Makefile.in
 config.log
 config.status
 src/backport-include/backport/backport_auto.h*
+src/backport-include/backport/automacro.h*
```
smuqthya, line 40: Does not say why we need to remove it or what is the significance of this file. why we added it in first place?

knayana-gse: Thanks for the review. 
automacro.h is a generated header created during modules build by create_automacro_h.sh, used for kernel-header-dependent trace compatibility macros, and is not a manually maintained source file.

This patch only updates .gitignore to avoid tracking that generated artifact, consistent with other generated headers already ignored.

Updated commit description as well.

smuqthya: Looks good to me. I asked the reverse question. but its answered

smuqthya (approved):

PR #310, opened 2026-03-26
drm/xe: Fix remap unwind overlap by tracking prev/next

kbommu:
During 3D workloads, a warning is triggered in the unwind
path due to VMA overlap.

This happens when rebind skips compatible ends, clears prev/next,
and shrinks the unmap range. On unwind, the ends remain while
the middle VMA is re-inserted, causing overlap.

Fix by keeping prev/next tracking and restoring the original
unmap range during unwind.

v2: Update message and restore unmap range for skip case.
v3: Track original start/range separately.


kbommu:
this changes are already part of https://github.com/intel-gpu/xekmd-backports/pull/311

PR #325, opened 2026-04-06
drm/xe: skip devcoredump registration on kernels without dev_coredump_put()

knayana-gse:
dev_coredump_put() was added in v6.10 (commit a28380f119a9). On older
kernels, xe_driver_devcoredump_fini() calling it unconditionally causes
mutex corruption (DEBUG_LOCKS_WARN_ON) during module unload/reload.

A header-only backport is not possible: dev_coredump_put() relies on
devcd_class, a file-scope static in drivers/base/devcoredump.c that is
never exported.

On kernels where BPM_DEVCOREDUMP_PUT_NOT_PRESENT is defined, the deferred
snapshot worker frees captured state immediately instead of registering a
persistent coredump entry, and the fini callback skips the
dev_coredump_put() call.

### DIFF
--- a/patches/0001-drm-xe-skip-devcoredump-registration-on-kernels-with.patch
+++ b/patches/0001-drm-xe-skip-devcoredump-registration-on-kernels-with.patch
@@ -1,0 +1,63 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Kanaka Raju Nayana <kanaka.raju.nayana@intel.com>
+Date: Mon, 6 Apr 2026 23:40:55 +0530
+Subject: drm/xe: skip devcoredump registration on kernels without
+ dev_coredump_put()
+
+dev_coredump_put() was added in v6.10 (commit a28380f119a9). On older
+kernels, xe_driver_devcoredump_fini() calling it unconditionally causes
+mutex corruption (DEBUG_LOCKS_WARN_ON) during module unload/reload.
+
+A header-only backport is not possible: dev_coredump_put() relies on
+devcd_class, a file-scope static in drivers/base/devcoredump.c that is
+never exported.
+
+On kernels below 6.10, where BPM_DEVCOREDUMP_PUT_NOT_PRESENT is defined,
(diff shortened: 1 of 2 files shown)


patches/0001-drm-xe-skip-devcoredump-registration-on-kernels-with.patch
```diff
@@ -0,0 +1,63 @@
...
+
+dev_coredump_put() was added in v6.10 (commit a28380f119a9). On older
+kernels, xe_driver_devcoredump_fini() calling it unconditionally causes
+mutex corruption (DEBUG_LOCKS_WARN_ON) during module unload/reload.
+
+A header-only backport is not possible: dev_coredump_put() relies on
+devcd_class, a file-scope static in drivers/base/devcoredump.c that is
+never exported.
+
+On kernels below 6.10, where BPM_DEVCOREDUMP_PUT_NOT_PRESENT is defined,
```
smuqthya, line 15: we should have it as a limitation in documentation. Please add a note on the readme

smuqthya: What is the implication of the calling freeing snapshot. are we going to miss coredump during issue

patches/0001-drm-xe-skip-devcoredump-registration-on-kernels-with.patch
```diff
@@ -0,0 +1,63 @@
...
+ 			      XE_COREDUMP_TIMEOUT_JIFFIES);
++#endif
+ 
+ 	xe_pm_runtime_get(xe);
+ 
+@@ -412,7 +421,9 @@ static void xe_driver_devcoredump_fini(void *arg)
+ {
+ 	struct drm_device *drm = arg;
+ 
++#ifndef BPM_DEVCOREDUMP_PUT_NOT_PRESENT
```
smuqthya, line 56: instead cant we simply pass NULL during init if macro is defined and that way , this function wont be called "xe_driver_devcoredump_fini"

PR #326, opened 2026-04-07
backport/base: Add stepping info and NVL-P workaround patches

pravalikagurram1:
drm/xe: Add NVL-P platform stepping support and related cleanups

Prepare the driver for NVL-P enablement by introducing support for
platform-level stepping and related infrastructure updates.

Modify xe_step_*_get() helpers to update xe->info.step directly in
preparation for adding platform-level stepping information to
struct xe_step_info. Add support for reading platform-level stepping
from PCI RevID on NVL-P systems and extend the RTP matching logic to
allow rules to match against platform-level stepping.

Implement the device workaround Wa_14026539277 for NVL-P A0 while
skipping application on SR-IOV VFs. Remove unused macros
IS_PLATFORM_STEP() and IS_SUBPLATFORM_STEP() that are no longer used
by the driver.

Drop the ineffective Wa_16028780921 implementation since the register
is locked by firmware before driver load and the workaround belongs
to firmware.

Finally, translate the C-state "reset value" returned on NVL-P systems
to RC6 when reading the GT sleep state, since higher-level sleep
states imply the GT is already in C6.


pravalikagurram1:
since change is merged already on internal so it cherry-pick only hence merging this without review

smuqthya (approved):

PR #347, opened 2026-04-20
handle unknown flavor parameters and features treat as default behavior

pravalikagurram1:
Previously, unknown flavor parameters (e.g., ./backport.sh -c xyz) would cause
a file not found error. Now they default to features behavior, applying all
patches except oot. This prevents script failures from typos or new flavor
options while maintaining explicit handling for base and oot.

Fixes: 6fd09f300 ("Implement selective patch application with flavor options")

### DIFF
--- a/backport.sh
+++ b/backport.sh
@@ -59,8 +59,13 @@ apply_patches() {
 					break
 				fi
 
-			elif [ -z "$flavor" ] || [[ "$flavor" == "features" ]]; then
-				# Default or features: Apply everything EXCEPT oot
+			elif [[ "$flavor" == "oot" ]]; then
+				# Apply everything including oot
+				echo $p | tr -d "#" | (read name; echo "Applying $name patches..!" >&2)
+				continue
+
+			else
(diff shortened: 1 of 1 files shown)


backport.sh
```diff
@@ -59,20 +59,20 @@ apply_patches() {
 					break
 				fi
 
-			elif [ -z "$flavor" ] || [[ "$flavor" == "features" ]]; then
-				# Default or features: Apply everything EXCEPT oot
+			elif [[ "$flavor" == "oot" ]]; then
```
smuqthya, line 62: It seems that we are trying to cater to any thing passed by user which are not given as option similar to default option. This seems to be bug in script. script should parse arguments and if any arguments are not part of options provided. we should throw back saying not supported. If none arguments are given, its considered default, which is fine.  If users diverting from the required option, it should be fix at their end. we should limit ourself to warning and exit.

smuqthya (changes requested):

pravalikagurram1:
closing this as this is not required

PR #355, opened 2026-04-22
Refactor: Use switch statement for flavor selection in apply_patches()

pravalikagurram1:
Convert nested if-elif-else logic to a switch/case statement for improved
readability. No functional changes - only refactoring for better code structure.

### DIFF
--- a/backport.sh
+++ b/backport.sh
@@ -47,45 +47,50 @@ apply_patches() {
 	echo "Applying local patches"
 	echo "Flavor: $flavor"
 	while read p; do
-		case "$p" in \#*)
+		if [[ "$p" =~ ^\# ]]; then
 			patch_category=$(echo $p | tr -d "#" | xargs)
-			if [[ "$flavor" == "base" ]]; then
-				# Apply only the base section
-				if [[ "$patch_category" == "base" ]]; then
+
+			case "$flavor" in
+				base)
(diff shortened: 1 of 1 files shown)


backport.sh
```diff
@@ -49,37 +49,41 @@ apply_patches() {
...
-			else
-				# Invalid flavor provided
-				echo "ERROR: Flavor '$flavor' is not listed"
-				echo "Supported flavors are: base, features (default), oot"
-				echo "Please check README for more information"
-				exit 1
-			fi
+					;;
+				*)
+					# Invalid flavor provided
```
smuqthya, line 80: Should point to usage instead. all usage data should be in the function - usage () {

smuqthya: Currently it does not have flavor data

smuqthya (changes requested):

backport.sh
```diff
@@ -50,37 +49,42 @@ apply_patches() {
 	while read p; do
 		case "$p" in \#*)
```
smuqthya: Do we need case for only 1 thing

smuqthya (approved):
Check if its need as case or move to if. reset is fine

backport.sh
pravalikagurram1: Done added if

PR #375, opened 2026-04-27
backport: Add error handling for unsupported flavor options

pravalikagurram1:
Add an else case to the flavor validation logic in apply_patches() to
handle invalid or unsupported flavor specifications. When a user
provides a flavor that is not in the supported list (base, features, oot),
the script now displays an error message with the list of supported options
and exits with error status code 1, instead of silently continuing."

### DIFF
--- a/backport.sh
+++ b/backport.sh
@@ -47,9 +47,11 @@ apply_patches() {
 	echo "Applying local patches"
 	echo "Flavor: $flavor"
 	while read p; do
-		case "$p" in \#*)
+		if [[ "$p" =~ ^\# ]]; then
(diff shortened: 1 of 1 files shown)


ayazahma (changes requested):
there is no point to have two commits , they should merge togather 
I think we should provide , also in case error , we should just print waring that un supporrted option  and fall back to default

ayazahma:
Also This change is not critical, so I suggest completing functional implementier oot backport  Change etc along with these Change 

can I have pr for backport tool, to use oot option
pr for oot-backport to remove patch from oot-backport and use kernel backport patch

smuqthya (approved):

PR #378, opened 2026-04-28
drm/xe: use dev_coredumpv() without dev_coredump_put()

knayana-gse:
Older kernels do not provide dev_coredump_put(). On that path, publishing
through dev_coredumpm(..., THIS_MODULE, ...) keeps xe-owned callbacks tied to
the devcoredump entry lifetime and can block module reload.

For BPM_DEVCOREDUMP_PUT_NOT_PRESENT, materialize the dump into a vmalloc
buffer and hand it to dev_coredumpv(). This keeps devcoredump visibility on
older kernels while removing xe callback lifetime from the old-kernel path.
Also compile out the legacy read/free helpers when they are not used.

### DIFF
--- a/patches/0004-drm-xe-use-dev_coredumpv-without-dev_coredump_put.patch
+++ b/patches/0004-drm-xe-use-dev_coredumpv-without-dev_coredump_put.patch
@@ -1,0 +1,110 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Kanaka Raju Nayana <kanaka.raju.nayana@intel.com>
+Date: Tue, 28 Apr 2026 23:58:46 +0530
+Subject: drm/xe: use dev_coredumpv() without dev_coredump_put()
+
+Below 6.10 kernels do not provide dev_coredump_put(). On that path,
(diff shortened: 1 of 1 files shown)


patches/0004-drm-xe-use-dev_coredumpv-without-dev_coredump_put.patch
```diff
@@ -0,0 +1,110 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Kanaka Raju Nayana <kanaka.raju.nayana@intel.com>
+Date: Tue, 28 Apr 2026 23:58:46 +0530
+Subject: [PATCH] drm/xe: use dev_coredumpv() without dev_coredump_put()
```
smuqthya: Remove [PATCH]

smuqthya (approved):

patches/0004-drm-xe-use-dev_coredumpv-without-dev_coredump_put.patch
knayana-gse: done

PR #394, opened 2026-05-12
drm/xe: Use feature flag for Xe2-style blitter instructions

kbommu:
The blitter MEM_COPY and MEM_SET instructions were introduced alongside
service copy engines (BCS1-BCS8), so the driver historically used
service copy engine presence to determine whether to use these
instructions instead of legacy XY_* commands.

However, this decision should be based on architectural support rather
than the set of currently usable engines. Platforms whose graphics
descriptor includes service copy engines support the newer instructions
even if all service copy engines are fused off. The existing logic broke
when gt->info.__engine_mask was removed, although no known SKU has
exposed the issue yet.

Replace has_service_copy_support() with a dedicated GT feature flag,
has_xe2_blt_instructions, that directly tracks support for the newer
blitter instructions. This also makes the intent of the calling code
clearer since the condition is no longer indirectly tied to service copy
engine availability.

The new flag is expected to be enabled for all Xe2+ platforms and PVC,
which also supports these instructions despite being Xe1-era hardware.

Fixes: 61549a2ee594 ("drm/xe: Drop _engine_mask")

### DIFF
--- a/backport/patches/fixes/base/0001-drm-xe-Make-decision-to-use-Xe2-style-blitter-instru.patch
+++ b/backport/patches/fixes/base/0001-drm-xe-Make-decision-to-use-Xe2-style-blitter-instru.patch
@@ -1,0 +1,130 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Matt Roper <matthew.d.roper@intel.com>
+Date: Thu, 7 May 2026 14:00:15 -0700
+Subject: drm/xe: Make decision to use Xe2-style blitter instructions a
+ feature flag
+
+The blitter engines' MEM_COPY and MEM_SET instructions were added as
+part of the same hardware change that introduced service copy engines
+(i.e., BCS1-BCS8) which is why the driver checks for service copy engine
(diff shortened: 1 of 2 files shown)


backport/patches/fixes/base/0001-drm-xe-Make-decision-to-use-Xe2-style-blitter-instru.patch
```diff
@@ -0,0 +1,130 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Matt Roper <matthew.d.roper@intel.com>
+Date: Thu, 7 May 2026 14:00:15 -0700
+Subject: [PATCH] drm/xe: Make decision to use Xe2-style blitter instructions a
```
pravalikagurram1: remove patch

kbommu: updated

pravalikagurram1 (approved):

PR #404, opened 2026-05-19
Set environment variables in make command

akanksh5-intc:
Instead of requiring a /usr/bin/yacc symlink to be manually created on the target system, export LEX=flex and YACC=bison directly in the dkms.conf MAKE and CLEAN commands. This allows bison to satisfy the yacc requirement without modifying system paths, removing the need for the manual steps.


knayana-gse (approved):
LGTM

PR #418, opened 2026-06-07
Rebase for 7.0 against supported kernels

akanksh5-intc:
### Description

**Summary:** 
Add a set of m4 probes and matching backport compatibility shims so this tree can build cleanly against the supported kernels.

- Add autotools probes to detect kernel APIs: access_remote_vm, EXPORT_SYMBOL_FOR_MODULES, kmap_local, bin_attribute callbacks, folio/pagemap helpers, etc.
- Backport compatibility implementations where appropriate (fallback access_remote_vm, drm_dev_wedged_event, small header shims in backport/backport-include/*).
- Adjust xe.m4 to use the new probes.


autotools/m4/drm_dumb_buffers.m4
```diff
@@ -5,12 +5,11 @@ dnl #
...
 	AC_KERNEL_DO_BACKGROUND([
 		AC_KERNEL_TRY_COMPILE([
-			#define DRM_FORMAT_INVALID	0
 			#include <drm/drm_dumb_buffers.h>
 		],[
-			struct drm_device *dev = NULL;
-			struct drm_mode_create_dumb *args = NULL;
-			drm_mode_size_dumb(dev, args, 0, 0);
+			drm_mode_size_dumb((struct drm_device *)0,
+					   (struct drm_mode_create_dumb *)0,
```
pravalikagurram1, line 11: Is it improvement. I dont see any correlation between struct changes and this?

backport/backport-include/linux/swap.h
```diff
@@ -4,21 +4,28 @@
...
+#ifdef BPM_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NOT_PRESENT
 static inline void *kmap_local_page_try_from_panic(struct page *page)
 {
-        return page_address(page);
+        return KMAP_SAFE_PAGE_ADDRESS(page);
 }
 #endif
+
+#ifdef BPM_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NONCONST_PRESENT
+static inline void *backport_kmap_const(const struct page *page)
```
pravalikagurram1, line 25: where we are using this?

backport/backport-include/linux/swap.h
```diff
@@ -4,21 +4,28 @@
...
-        return NULL;
-}
+#define KMAP_SAFE_PAGE_ADDRESS(page) \
+        (PageHighMem(page) ? NULL : page_address(page))
 #else
+#define KMAP_SAFE_PAGE_ADDRESS(page) \
+        page_address(page)
+#endif
+
+#ifdef BPM_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NOT_PRESENT
```
pravalikagurram1, line 17: I do see error is different and the change is differnet 
I see error is speaking about API arg but here it handled in differnet way

patches/0001-autotools-m4-Manage-the-availability-of-const-bin_at.patch
```diff
@@ -0,0 +1,79 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Akanksha Hotta <akanksha.hotta@intel.com>
+Date: Tue, 9 Jun 2026 10:47:10 +0530
+Subject: autotools/m4: Manage the availability of const bin_attribute
+ pointer
+
+Use BPM_CONST_STRUCT_BIN_ATTRIBUTE_IS_NOT_PRESENT (not
+BPM_STRUCT_BIN_ATTRIBUTE_READ_NEW_NOT_PRESENT) to select the const
+vs non-const bin_attribute pointer in intel_pmt_read() and
+intel_pmt_mmap() signatures. On kernel 7.0 the transitional
```
pravalikagurram1: what is the difference between this and deleted patch patches/0001-pmt-Fix-bin_attribute-read-member-name-compatibility.patch

backport/compat/drm_drv.c
```diff
@@ -54,7 +54,8 @@
 #include <drm/drm_privacy_screen_machine.h>
 
 
-#ifdef BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT
+#if defined(BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT) && \
+    defined(BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT)
```
pravalikagurram1: I see error is speaking that api exported twice but the fix is about Args how both are correlated ?

patches/0001-autotools-m4-Manage-the-availability-of-const-bin_at.patch
akanksh5-intc: It's not using the macro: BPM_STRUCT_BIN_ATTRIBUTE_READ_NEW_NOT_PRESENT instead now it's using `BPM_CONST_STRUCT_BIN_ATTRIBUTE_IS_NOT_PRESENT`.


Reason of the change:
     7ff2fecc8bc2e85
      platform/x86/intel/pmt: Constify 'struct bin_attribute'

In kernels that are below v6.14 the below was there, but it got updated in future kernel:

**Function Signature Update:**
- Changes the intel_pmt_read() function parameter from struct bin_attribute *attr to const struct bin_attribute *attr
- This makes the attribute pointer read-only within the function
**Attribute Assignment Change:**
- Changes from entry->pmt_bin_attr.read = intel_pmt_read;
to entry->pmt_bin_attr.read_new = intel_pmt_read;

autotools/m4/drm_dumb_buffers.m4
akanksh5-intc: Actually, I got this error while compiling against 7.0 headers
`ERROR: modpost: "drm_mode_size_dumb" [/home/psb-backport/7.0-changes/pr_changes/out-dir-final/src/drivers/gpu/drm/xe/xe.ko] undefined!`

This macro BPM_DRM_MODE_SIZE_DUMB_NOT_PRESENT was defined while compiling against 7.0 headers, expected is it shouldn't define why because the compat block compile and EXPORT_SYMBOL(drm_mode_size_dumb).

So, if that macro compile tests in m4 macros run with -nostdinc (no standard C headers). NULL is defined in <stddef.h>, which is a standard header so under -nostdinc it is simply undefined.

backport/backport-include/linux/swap.h
akanksh5-intc: Probably my commit message didn't justify to it. Thanks for pointing it out. So, I got the below error

```
/home/psb-backport/7.0-changes/pr_changes/out-dir-final/src/backport-include/linux/swap.h:17:21: error: conflicting types for ‘kmap_local_page_try_from_panic’; have ‘void *(struct page *)’
17 | static inline void *kmap_local_page_try_from_panic(struct page *page)
| ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/psb-backport/xe-dkms/drivers.gpu.kernel.backporting.headers/include/linux/highmem.h:14,
from drivers/gpu/drm/ttm/ttm_pool.c:38:
/home/psb-backport/xe-dkms/drivers.gpu.kernel.backporting.headers/include/linux/highmem-internal.h:191:21: note: previous definition of ‘kmap_local_page_try_from_panic’ with type ‘void *(const struct page *)’
191 | static inline void *kmap_local_page_try_from_panic(const struct page *page)
| ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The root cause: on kernel 7.0, highmem-internal.h can no longer be included directly in compile tests (so BPM_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NOT_PRESENT is wrongly defined in swap.h) tries to redefine the function with struct page * (non-const), which conflicts with the kernel's const struct page * definition.

To resolve that used <linux/highmem.h> header in the m4. 

But thanks for pointing it out, I will justified commit message for this change.

backport/backport-include/linux/swap.h
akanksh5-intc: The justification that is provided above hopefully helps now.

autotools/m4/export_symbol_for_modules.m4
```diff
@@ -6,12 +6,15 @@ AC_DEFUN([AC_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT], [
...
 		],[
-			EXPORT_SYMBOL_FOR_MODULES(sym, mods);
 		],[
 		],[
 			AC_DEFINE(BPM_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT, 1,
 				[EXPORT_SYMBOL_FOR_MODULES macro is not avilable])
 		])
 	])
 ])
+
```
smuqthya: unnecessary line

backport/backport-include/linux/export.h
```diff
@@ -14,7 +14,7 @@
 #define EXPORT_SYMBOL_NS_GPL(sym, ns) EXPORT_SYMBOL_GPL(sym)
 #endif
 
-#ifdef BPM_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT
+#if defined(BPM_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT) && !defined(EXPORT_SYMBOL_FOR_MODULES)
```
smuqthya: Do we  need M4 for this?

backport/backport-include/drm/drm_device.h
```diff
@@ -18,6 +18,7 @@
 
 #if defined(BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT) || \
         defined(BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT)
+#ifdef BPM_DRM_WEDGE_TASK_INFO_NOT_PRESENT
```
smuqthya, line 19: 3 diff macros controlling same code, is wrong.

smuqthya: please check and understand the changes and work on control of code.

autotools/m4/xe.m4
```diff
@@ -78,6 +78,7 @@ AC_DEFUN([AC_XE_CONFIG], [
 	AC_CLEANUP_ACQUIRE_NOT_PRESENT
 	AC_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT
 	AC_DRM_WEDGE_TASK_INFO_NOT_PRESENT
+	AC_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NOT_PRESENT
```
smuqthya: why adding already added one

backport/backport-include/linux/swap.h
```diff
@@ -4,21 +4,28 @@
...
+/* Helper macro to avoid duplicating the HighMem guard logic */
 #ifdef CONFIG_HIGHMEM
-static inline void *kmap_local_page_try_from_panic(struct page *page)
-{
-        if (!PageHighMem(page))
-                return page_address(page);
-        /* If the page is in HighMem, it's not safe to kmap it.*/
-        return NULL;
-}
+#define KMAP_SAFE_PAGE_ADDRESS(page) \
```
smuqthya, line 9: better to add ifndef in case of macro definitions

autotools/m4/bin_attribute.m4
```diff
@@ -31,8 +31,13 @@ AC_DEFUN([AC_STRUCT_BIN_ATTRIBUTE_READ_NEW_NOT_PRESENT], [
                 AC_KERNEL_TRY_COMPILE([
                         #include <linux/sysfs.h>
                 ],[
-                        struct bin_attribute test_attr;
-                        test_attr.read_new = NULL;
+                        ssize_t (*test_read_new)(struct file *, struct kobject *,
+                                                 const struct bin_attribute *,
```
smuqthya, line 35: If we are updating m4 from non const to const, isnt it suppose to be other m4 or nested

smuqthya: Did we compile on the other OSV?

patches/0001-autotools-m4-Manage-the-availability-of-const-bin_at.patch
```diff
@@ -0,0 +1,79 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Akanksha Hotta <akanksha.hotta@intel.com>
+Date: Tue, 9 Jun 2026 10:47:10 +0530
+Subject: autotools/m4: Manage the availability of const bin_attribute
```
smuqthya: Patches should not have m4 title

backport/compat/drm_drv.c
smuqthya: how are we controlling code with both macros, where is the alternative.

autotools/m4/access_remote_vm.m4
```diff
@@ -17,6 +18,8 @@ AC_DEFUN([AC_ACCESS_REMOTE_VM_NOT_PRESENT], [
 			func_ptr = access_remote_vm;
 			(void)func_ptr;
 		],[
+			AC_DEFINE(BPM_ACCESS_REMOTE_VM_EXPORT_NOT_PRESENT, 1,
```
smuqthya: what are we trying to achieve here?

smuqthya: why macro defined at both cases, only one need to be verified

backport/compat/memory.c
```diff
@@ -41,30 +41,27 @@
 
 #include <linux/mm.h>
 
-#ifdef BPM_ACCESS_REMOTE_VM_NOT_PRESENT
+#if defined(BPM_ACCESS_REMOTE_VM_NOT_PRESENT) || \
```
smuqthya: compile on old

smuqthya (changes requested):

backport/compat/memory.c
akanksh5-intc: When I compiled against old its compiling with 6.6,6.8 and rhel_10.1 headers. But while compiling against 7.0 headers I am getting 

ERROR: modpost: "access_remote_vm" [drivers/gpu/drm/xe/xe.ko] undefined!

So why I created another macro like `BPM_ACCESS_REMOTE_VM_EXPORT_NOT_PRESENT` nested to `AC_ACCESS_REMOTE_VM_NOT_PRESENT`, because once the macro fails to define against 7.0 headers this macro (`BPM_ACCESS_REMOTE_VM_EXPORT_NOT_PRESENT` can be set to 1) to resolve the modpost found the symbol from xe-compat.

autotools/m4/export_symbol_for_modules.m4
akanksh5-intc: Addressed.

backport/backport-include/linux/export.h
akanksh5-intc: I removed the usage of macro instead handling like

```
#ifndef EXPORT_SYMBOL_FOR_MODULES
#define EXPORT_SYMBOL_FOR_MODULES(sym, mods) __EXPORT_SYMBOL(sym, "GPL", mods)
#endif
```

autotools/m4/xe.m4
akanksh5-intc: Removed.

backport/backport-include/linux/swap.h
akanksh5-intc: Yes addressed it like `#ifndef KMAP_SAFE_PAGE_ADDRESS`

patches/0001-autotools-m4-Manage-the-availability-of-const-bin_at.patch
akanksh5-intc: Removed now

autotools/m4/access_remote_vm.m4
akanksh5-intc: > why macro defined at both cases, only one need to be verified

This I replied the reason below.

akanksh5-intc: > what are we trying to achieve here?

Need to resolve the modpost error while compiling against 7.0 headers.

autotools/m4/bin_attribute.m4
akanksh5-intc: > Did we compile on the other OSV?

Yes against all OSV's its working fine.

backport/compat/drm_drv.c
akanksh5-intc: `BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT` - kernel has no drm_dev_wedged_event symbol (checked with 2-arg probe).

`BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT` - kernel does not support the 3rd argument (checked with 3-arg probe).

- That block is the full backport implementation of drm_dev_wedged_event. It should compile only when the symbol is truly absent, which is represented here by both macros being true.

akanksh5-intc: `BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT` - kernel has no drm_dev_wedged_event symbol (checked with 2-arg probe).

`BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT` - kernel does not support the 3rd argument (checked with 3-arg probe).

- That block is the full backport implementation of drm_dev_wedged_event. It should compile only when the symbol is truly absent, which is represented here by both macros being true.

backport/backport-include/linux/export.h
```diff
@@ -14,7 +14,7 @@
 #define EXPORT_SYMBOL_NS_GPL(sym, ns) EXPORT_SYMBOL_GPL(sym)
 #endif
 
-#ifdef BPM_EXPORT_SYMBOL_FOR_MODULES_NOT_PRESENT
+#ifndef EXPORT_SYMBOL_FOR_MODULES
```
smuqthya, line 17: M4 removed?

smuqthya: commit message is for m4. change it

backport/backport-include/drm/drm_device.h
```diff
@@ -28,6 +27,7 @@ struct drm_wedge_task_info {
 	char comm[TASK_COMM_LEN];
 };
 #endif
+#endif
```
smuqthya: remove

backport/backport-include/linux/swap.h
```diff
@@ -4,21 +4,29 @@
...
 #endif
+
+#ifdef BPM_KMAP_LOCAL_PAGE_TRY_FROM_PANIC_NONCONST_PRESENT
+static inline void *backport_kmap_const(const struct page *page)
+{
+        return KMAP_SAFE_PAGE_ADDRESS(page);
+}
 #endif
 
+
```
smuqthya: unnecessary new line

patches/0001-autotools-m4-Manage-the-availability-of-const-bin_at.patch
```diff
@@ -0,0 +1,78 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
smuqthya: replace the patch instead of removing original owner of change

backport/compat/drm_drv.c
smuqthya: M4 should be part of the same change.

autotools/m4/access_remote_vm.m4
```diff
@@ -17,6 +18,8 @@ AC_DEFUN([AC_ACCESS_REMOTE_VM_NOT_PRESENT], [
 			func_ptr = access_remote_vm;
 			(void)func_ptr;
```
smuqthya: Why not just call function. I need to check how "(void)func_ptr; " this is behaving

autotools/m4/access_remote_vm.m4
smuqthya: Un necessary variables

backport/backport-include/linux/mm.h
```diff
@@ -0,0 +1,13 @@
+/* SPDX-License-Identifier: MIT */
+#ifndef __BACKPORT_LINUX_MM_H
+#define __BACKPORT_LINUX_MM_H
+
+#include_next <linux/mm.h>
+
+#if defined(BPM_ACCESS_REMOTE_VM_NOT_PRESENT) || \
```
smuqthya: we are just covering both pass and fail cases

smuqthya (changes requested):

backport/backport-include/linux/export.h
akanksh5-intc: No, the M4 was not removed it's there as it is (export_symbol_for_modules.m4) as it's been utilized in vfio/pci/xe/main.c. 

So just with latest changes now instead of using the macro just checking if the api is undefined then define it, else if its defined somewhere then it will skip.

backport/compat/drm_drv.c
akanksh5-intc: I wrongly told on discussion that for this macro BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT, I created the function, but it was already present in autotools/m4/drm_dev_wedged.m4. I have utilized it.

autotools/m4/access_remote_vm.m4
akanksh5-intc: Removed now.

backport/backport-include/linux/mm.h
akanksh5-intc: Addressed it now properly. And also checked it behaviors against all supported osv's it is working fine.

autotools/m4/access_remote_vm.m4
akanksh5-intc: It simply suppresses the variable set but not used to eradicate compiler warning (-Wunused-variable). Without it, it will warn that func_ptr is assigned but never read.

And as we are suppressing the variables to null-ops so removed those unnecessary variables which are assigned with NULL/zero. (Ref: https://github.com/intel-gpu/xekmd-backports/pull/418/changes/f298ae1118c69b4dca0bc225bf02de0cd114ff7b#diff-a269d25711271cd7b801e072cf0f1c9157ce24452bab5d83a93298c76412382b)

backport/backport-include/linux/export.h
smuqthya: if we dont need the M4 here, that mean we may not need M4 there as well. we should have a clean-up Jira as part of the seperate task if it takes time.

backport/compat/drm_drv.c
```diff
@@ -54,7 +54,8 @@
 #include <drm/drm_privacy_screen_machine.h>
 
 
-#ifdef BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT
+#if defined(BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT) && \
```
smuqthya: Isnt this wrong as per the discussion

backport/compat/drm_drv.c
smuqthya: I remembered. Thanks

backport/compat/drm_drv.c
smuqthya: && is not the proper here. check based on the the sequence of additions

akanksh5-intc: image


`Below 6.15KV (function absent):`
  → BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT defined
  → BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT defined
  → compat.c #ifdef NOT_PRESENT  ← provides full 3-arg implementation

`On 6.15KV (2-arg function added):`
  → BPM_DRM_DEV_WEDGED_EVENT_NOT_PRESENT  **not-defined**
  → BPM_DRM_DEV_WEDGED_EVENT_ARG3_NOT_PRESENT **defined**
  → drm_drv.h provides → 2-arg wrapper
  
`6.16+KV (3rd arg added):`
  → neither macro is defined
  → kernel provides full 3-arg function, no compat needed

smuqthya (approved):

smuqthya:
conditional. results pending

akanksh5-intc:
> conditional. results pending

BAT results: https://gtax-ril-fm.intel.com/#/jobs/4845206

PR #422, opened 2026-06-08
Update README.md with branches 7.0

smuqthya:
Update README.md with branches of 7.0 oot and releases


ayazahma (changes requested):
1. v7.0 branch is for 7.0 only, no plus
2. There is not OOt 7.0 as well as release 7.0, so why we are adding here

ayazahma (approved):

PR #426, opened 2026-06-11
backport/Makefile: Add 7.0KV build support and fix deb-pkg versioning

akanksh5-intc:


autotools/Makefile.am
```diff
@@ -59,8 +59,8 @@ DISTRO_SUFFIX=@DOLLAR_SIGN@(shell cat $(KLIB_BUILD)/include/generated/utsrelease
 XE_PKG_NAME_BASENAME=intel-xe-dkms
```
smuqthya, line 59: Lets add previous package name and version info and new package name and version after this change

smuqthya (approved):
Conditional

autotools/Makefile.am
akanksh5-intc: Added on the same file.

PR #419, opened 2026-06-08
XE-KMD Backport-Release: xebr-v6.17.13.62.260409.9

seos-autobackport:
Generation Information:
XeKMD-Backport-Tag: v6.17.13.62
Backports Tag: xeb_v6.17.13.62_260409.9
Backports Branch: oot-backport/main
Backports Head: fcb7c5dbe

Signed-off-by: SYS SEOS AUTOBACKPORT <sys_seos_autobackport@intel.com>

PR #432, opened 2026-06-15
Revert "XE-KMD Backport-Release: xebr-v6.17.13.62.260409.9"

akanksh5-intc:
Reverts intel-gpu/xekmd-backports#419


src/versions
```diff
@@ -1,3 +1,3 @@
-BACKPORTS_RELEASE_TAG="xeb_v6.17.13.62_260409.9"
+BACKPORTS_RELEASE_TAG="xeb_v6.17.13.62_260409.8"
```
smuqthya, line 1: version should increment in reverts

smuqthya (changes requested):

src/versions
smuqthya: Should be release from oot-backorot

PR #444, opened 2026-06-29
xekmd-backport: backport CRI patches on top of v7.1.1

kbommu:
Backport CRI (Crescent Island) platform enablement patches on top of
the v7.1.1 Intel GPU kernel baseline. The series includes platform
bring-up (xe_sysctrl, PCI IDs, GuC firmware), RAS correctable error
handling, memirq updates, OA workarounds, and GT/hwmon fixes.


series
```diff
@@ -1,5 +1,44 @@
 # base
 backport/patches/base/0001-drm-xe-tuning-Stop-applying-CCCHKNREG1-tuning-from-X.patch
+# features/cri
```
smuqthya: cri is not a feature. This should land in base if not related to particular. or we should have new implementation for platforms, but I dont see much use as of now for separate platform section.

smuqthya (changes requested):

series
kbommu: ok I will add in base and update

smuqthya (approved):

PR #451, opened 2026-07-06
xekmd-backports: Backport 7.1 kernel for supported osv's

akanksh5-intc:


backport/backport-include/linux/dma-buf.h
```diff
@@ -21,4 +21,12 @@ MODULE_IMPORT_NS(DMA_BUF);
 MODULE_IMPORT_NS("DMA_BUF");
 #endif
 
+#if !BPM_DMA_BUF_ATTACH_OPS_INVALIDATE_MAPPINGS_NOT_PRESENT
```
pravalikagurram1: diff --git a/src/backport-include/linux/dma-buf.h b/src/backport-include/linux/dma-buf.h
index 648f680..1571b9e 100644
--- a/src/backport-include/linux/dma-buf.h
+++ b/src/backport-include/linux/dma-buf.h
@@ -28,4 +28,14 @@ static inline void dma_buf_invalidate_mappings(struct dma_buf *dmabuf)
 }
 #endif

+#ifdef BPM_DMA_BUF_ATTACH_OPS_INVALIDATE_MAPPINGS
+/*
+ * Kernel 6.6: dma_buf_attach_ops uses 'move_notify' member
+ * This macro allows using 'invalidate_mappings' in struct initializers,
+ * which gets mapped to 'move_notify' for 6.6 compatibility.
+ */
+#define invalidate_mappings move_notify
+#endif
+/* Kernel 6.8+: invalidate_mappings member is already present, no mapping needed */

I did like this Akanksha can you cross check once

backport/backport-include/linux/dma-fence.h
```diff
@@ -19,4 +19,15 @@ dma_fence_check_and_signal_locked(struct dma_fence *fence)
 }
 #endif /* BPM_DMA_FENCE_CHECK_AND_SIGNAL_LOCKED_NOT_PRESENT */
 
+#ifdef BPM_DMA_FENCE_LOCK_IRQSAVE_AND_IRQRESTORE_NOT_PRESENT
+#define dma_fence_lock_irqsave(fence, flags) \
+	spin_lock_irqsave((fence)->lock, flags)
```
pravalikagurram1, line 28: I have single commit for this. but here its clubbed check with Muqthyar if it okay

backport/backport-include/linux/gpu_buddy.h
```diff
@@ -0,0 +1,43 @@
...
+/* Type mappings */
+#define gpu_buddy drm_buddy
+#define gpu_buddy_block drm_buddy_block
+#define gpu_buddy_mm drm_buddy_mm
+#define gpu_buddy_addr_state drm_buddy_addr_state
+
+/* Flag mappings */
+#if 0
+#define GPU_BUDDY_TOPDOWN_ALLOCATION  DRM_BUDDY_TOPDOWN_ALLOCATION
+#define GPU_BUDDY_CONTIGUOUS_ALLOCATION  DRM_BUDDY_CONTIGUOUS_ALLOCATION
```
pravalikagurram1: Remove if 0

pravalikagurram1: but how your handling thsi for 6.6 kernels. i have these change for older kernels and few more changes in .c files
commit f166340a919707e4233b1d97eda846b93c8a4e99
Author: Pravalika Gurram <pravalika.gurram@intel.com>
Date:   Mon Jun 15 15:09:04 2026 +0530

     This use older flags need to recheck

diff --git a/src/drivers/gpu/drm/xe/xe_ttm_vram_mgr.c b/src/drivers/gpu/drm/xe/xe_ttm_vram_mgr.c
index 5fd0d55..86ef247 100644
--- a/src/drivers/gpu/drm/xe/xe_ttm_vram_mgr.c
+++ b/src/drivers/gpu/drm/xe/xe_ttm_vram_mgr.c
@@ -80,13 +80,13 @@ static int xe_ttm_vram_mgr_new(struct ttm_resource_manager *man,
        INIT_LIST_HEAD(&vres->blocks);

        if (place->flags & TTM_PL_FLAG_TOPDOWN)
-               vres->flags |= GPU_BUDDY_TOPDOWN_ALLOCATION;
-
+               vres->flags |= DRM_BUDDY_TOPDOWN_ALLOCATION;
+#if 0
        if (place->flags & TTM_PL_FLAG_CONTIGUOUS)
-               vres->flags |= GPU_BUDDY_CONTIGUOUS_ALLOCATION;
-
+               vres->flags |= DRM_BUDDY_CONTIGUOUS_ALLOCATION;
+#endif
        if (place->fpfn || lpfn != man->size >> PAGE_SHIFT)
-               vres->flags |= GPU_BUDDY_RANGE_ALLOCATION;
+               vres->flags |= DRM_BUDDY_RANGE_ALLOCATION;

patches/0001-drm-xe-firmware-signature-failure-plus-vram-mgr-follow-up.patch
```diff
@@ -0,0 +1,62 @@
...
++		vres->flags |= DRM_BUDDY_CONTIGUOUS_ALLOCATION;
++#endif
+ 	if (place->fpfn || lpfn != man->size >> PAGE_SHIFT)
+-		vres->flags |= GPU_BUDDY_RANGE_ALLOCATION;
++		vres->flags |= DRM_BUDDY_RANGE_ALLOCATION;
+ 
+ 	if (WARN_ON(!vres->base.size)) {
+ 		err = -EINVAL;
+@@ -114,11 +114,27 @@ static int xe_ttm_vram_mgr_new(struct ttm_resource_manager *man,
+ 		goto error_unlock;
```
pravalikagurram1, line 35: This needs cleanup

akanksh5-intc: Addressed with your patch that you shared.

patches/0001-drm-xe-xe_svm-mmu-notifier-single-pass.patch
```diff
@@ -0,0 +1,50 @@
...
+ 	struct xe_device *xe = vm->xe;
+ 	struct drm_gpusvm_range *r, *first;
+ 	struct xe_tile *tile;
+@@ -278,10 +282,15 @@
+ 
+ 	xe_device_wmb(xe);
+ 
++	#ifdef HAVE_MMU_INTERVAL_NOTIFIER_TWOPASS
+ 	err = xe_tlb_inval_range_tilemask_submit(xe, vm->usm.asid, adj_start, adj_end,
+ 						 tile_mask, &batch);
```
akanksh5-intc: Done.

backport/backport-include/linux/gpu_buddy.h
akanksh5-intc: Addressed.

backport/backport-include/linux/dma-fence.h
akanksh5-intc: Ok sure.

backport/backport-include/linux/dma-buf.h
akanksh5-intc: Addressed.

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
```diff
@@ -9,12 +9,14 @@ register_shrinker() to shrinker_alloc()/shrinker_register()/shrinker_free()
 in newer kernels.
 Use m4 macro to detect API availability and provide compatibility.
```
ayazahma, line 10: On 7.1KV backport builds, ttm pool shrinker setup fails when the code
unconditionally uses the newer shrinker allocation flow introduced in newer
upstream kernels.

Why do we need to mention 7.1 , its no longer tagged with kernel version but api changes or non coamptiblity

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
```diff
@@ -9,12 +9,14 @@ register_shrinker() to shrinker_alloc()/shrinker_register()/shrinker_free()
...
 v2:
 Fix Hunk failure
 
 Reference:
     d35b5c98c1f1 drm/ttm: dynamically allocate the drm-ttm_pool shrinker
 
+Signed-off-by: Akanksha Hotta <akanksha.hotta@intel.com>
 Signed-off-by: Kanaka Raju Nayana <kanaka.raju.nayana@intel.com>
 Signed-off-by: Rachapalli Bandi Jagadeesh <jagadeesh.rachapalli.bandi@intel.com>
 ---
```
ayazahma, line 21: latest sign off should be at the end

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
```diff
@@ -48,19 +50,23 @@ index baf27c7..8440f27 100644
...
+ 	for_each_node(nid) {
+ 		sc.nid = nid;
 +#ifdef BPM_SHRINKER_ALLOC_NOT_PRESENT
-+	count = ttm_pool_shrinker_count(&mm_shrinker, &sc);
-+	seq_printf(m, "%lu/%lu\n", count,
-+		   ttm_pool_shrinker_scan(&mm_shrinker, &sc));
++		count = ttm_pool_shrinker_count(&mm_shrinker, &sc);
++		seq_printf(m, "%d: %lu/%lu\n", nid, count,
++			   ttm_pool_shrinker_scan(&mm_shrinker, &sc));
 +#else
```
ayazahma, line 60: noise

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
```diff
@@ -48,19 +50,23 @@ index baf27c7..8440f27 100644
...
- 	count = ttm_pool_shrinker_count(mm_shrinker, &sc);
- 	seq_printf(m, "%lu/%lu\n", count,
- 		   ttm_pool_shrinker_scan(mm_shrinker, &sc));
+ 		count = ttm_pool_shrinker_count(mm_shrinker, &sc);
+ 		seq_printf(m, "%d: %lu/%lu\n", nid, count,
+ 			   ttm_pool_shrinker_scan(mm_shrinker, &sc));
 +#endif
+ 	}
```
ayazahma, line 65: Noises

patches/0002-Add-debug-print-and-BACKPORT_MOD_VERSIONS-for-oot-mo.patch
```diff
@@ -9,6 +9,9 @@ the backport driver.
 Add BACKPORT_MOD_VERSIONS to display information
```
ayazahma, line 9: All hunk failure should be part of single commit

autotools/m4/bitmap_weighted.m4
```diff
@@ -0,0 +1,19 @@
+dnl #
```
ayazahma, line 1: autotools/backport: Manage bitmap_weighted_or() compat shim for older kernels

autotools/backport: Provide a fallback implementation for bitmap_weighted_or()

backport/backport-include/linux/dma-buf.h
```diff
@@ -21,4 +21,12 @@ MODULE_IMPORT_NS(DMA_BUF);
...
+/*
+ * In kernels before the rename, dma_buf_attach_ops used 'move_notify'
+ * instead of 'invalidate_mappings'. Map the new name to the old one.
+ */
+#define invalidate_mappings move_notify
+#endif
```
ayazahma: Against the backbord philosophy, to chenge the coded based on macro is not defined

ayazahma: look like two commits , bith need to be squashed to gather

autotools/m4/dma_buf.m4
```diff
@@ -0,0 +1,29 @@
...
+        )],
+        [
+            AC_DEFINE(BPM_DMA_BUF_ATTACH_OPS_INVALIDATE_MAPPINGS_NOT_PRESENT, 1,
+                [dma_buf_attach_ops has invalidate_mappings member])
+        ],
+        [
+            AC_DEFINE(BPM_DMA_BUF_ATTACH_OPS_INVALIDATE_MAPPINGS_NOT_PRESENT, 0,
+                [dma_buf_attach_ops has move_notify member instead])
+        ]
+    )
```
ayazahma, line 53: why we need both true and false ?

autotools/m4/dma_fence_lock_irqsave.m4
```diff
@@ -0,0 +1,20 @@
+dnl #
+dnl #
+dnl #
+dnl #
```
ayazahma, line 25: why no ref here

autotools/m4/dma_fence_lock_irqsave.m4
```diff
@@ -0,0 +1,20 @@
...
+			struct dma_fence *fence = NULL;
+			unsigned long flags;
+			dma_fence_lock_irqsave(fence, flags);
+			dma_fence_unlock_irqrestore(fence, flags);
+		],[
+		],[
+			AC_DEFINE([BPM_DMA_FENCE_LOCK_IRQSAVE_AND_IRQRESTORE_NOT_PRESENT], 1,
+				[dma_fence_lock_irqsave() and dma_fence_unlock_irqrestore() are not available])
+		])
+	])
```
ayazahma, line 40: does both dma_fence_lock_irqsave and dma_fence_unlock_irqrestore introduced by same commit

backport/backport-include/linux/dma-fence.h
```diff
@@ -19,4 +19,15 @@ dma_fence_check_and_signal_locked(struct dma_fence *fence)
...
+#define dma_fence_unlock_irqrestore(fence, flags) \
+	spin_unlock_irqrestore((fence)->lock, flags)
+#endif /* BPM_DMA_FENCE_LOCK_IRQSAVE_AND_IRQRESTORE_NOT_PRESENT */
+
+#ifndef BPM_DMA_FENCE_EXTERN_LOCK_PRESENT
+#define extern_lock lock
+#endif   /* BPM_DMA_FENCE_EXTERN_LOCK_PRESENT */
+
```
ayazahma: does the order is correct, lin29 has extern declaration and line 24 is using this

patches/0001-drm-gpusvm-disable-thp-migration-on-non-folio-kernels.patch
```diff
@@ -0,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Pravalika Gurram <pravalika.gurram@intel.com>
+Date: Sun, 21 Jun 2026 20:06:01 +0530
+Subject: drm/gpusvm: disable THP migration on non-folio kernels
+
+Signed-off-by: Pravalika Gurram <pravalika.gurram@intel.com>
+---
```
ayazahma: what is source of this patch, I do need any cherry-pick /backported reference 

Subject: drm/gpusvm: disable THP migration on non-folio kernels
do we keep "Subject"

patches/0001-drm-gpusvm-disable-thp-migration-on-non-folio-kernels.patch
```diff
@@ -0,0 +1,44 @@
...
+diff --git a/drivers/gpu/drm/drm_gpusvm.c b/drivers/gpu/drm/drm_gpusvm.c
+index 4b928fd..68dcbf3 100644
+--- a/drivers/gpu/drm/drm_gpusvm.c
++++ b/drivers/gpu/drm/drm_gpusvm.c
+@@ -1488,15 +1488,25 @@ map_pages:
+ 		order = drm_gpusvm_hmm_pfn_to_order(pfns[i], i, npages);
+ 		if (is_device_private_page(page) ||
+ 		    is_device_coherent_page(page)) {
++#ifndef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
+ 			struct drm_pagemap_zdd *__zdd =
```
ayazahma: lINE 19 : 
why we have a code with NOT MARCO, there should be no change in case of macro is not defined, it Schuld be equivalent to original code of base kernel.

patches/0001-drm-pagemap-compat-for-older-kernels.patch
```diff
@@ -0,0 +1,192 @@
...
+ 		drm_pagemap_get_devmem_page(page, order, zdd);
+ 
+@@ -791,7 +808,11 @@
+ 
+ 		/* TODO: Support fallback to single pages if THP allocation fails */
+ 		if (vas)
++#ifdef BPM_VMA_ALLOC_FOLIO_4ARGS_NOT_PRESENT
++			folio = vma_alloc_folio(GFP_HIGHUSER, order, vas, addr, false);
++#else
+ 			folio = vma_alloc_folio(GFP_HIGHUSER, order, vas, addr);
```
pravalikagurram1: Remove this from THP patch . And it in .h file

patches/0001-drm-pagemap-compat-for-older-kernels.patch
```diff
@@ -0,0 +1,192 @@
...
+ 			mpfn[i] |= MIGRATE_PFN_COMPOUND;
++#endif
+ next:
+ 		if (page)
+ 			addr += page_size(page);
+@@ -1042,10 +1065,14 @@
+ 	if (err)
+ 		goto err_free;
+ 
++#ifdef BPM_MIGRATE_DEVICE_PFNS_NOT_PRESENT
```
pravalikagurram1: Remove this as well not part of this patch scope

patches/0001-drm-pagemap-compat-for-older-kernels.patch
```diff
@@ -0,0 +1,192 @@
...
++#endif
+ 		.fault_page	= page,
+ 	};
+ 	struct drm_pagemap_migrate_details mdetails = {};
+@@ -1222,12 +1254,19 @@
+  * This function is a callback used to put the GPU SVM zone device data
+  * associated with a page when it is being released.
+  */
++#ifdef BPM_DEV_PAGEMAP_OPS_FOLIO_FREE_NOT_PRESENT
++static void drm_pagemap_folio_free(struct page *page)
```
pravalikagurram1: We should not nave have any other macro in this patch

other than BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
so please check weather this code can be under BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT

patches/0001-drm-gpusvm-disable-thp-migration-on-non-folio-kernels.patch
```diff
@@ -0,0 +1,44 @@
...
+ 				err = -EOPNOTSUPP;
+ 				goto err_unmap;
+ 			}
++#ifdef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
++			zdd = page->zone_device_data;
++#else
+ 			zdd = __zdd;
++#endif
+ 			if (pagemap != page_pgmap(page)) {
+ 				if (pagemap) {
```
pravalikagurram1: Do the clean up in first commit
then raise changes for THP

patches/0001-drm-gpusvm-disable-thp-migration-on-non-folio-kernels.patch
```diff
@@ -0,0 +1,44 @@
...
+ 					err = -EOPNOTSUPP;
+--
```
ayazahma: just need to overwrite __zdd based on condition

patches/0001-drm-pagemap-Disable-THP-migration-fall-back-to-page-.patch
```diff
@@ -0,0 +1,192 @@
...
++#endif
+ }
+ 
+ /**
+@@ -455,6 +459,7 @@
+ 	return ret;
+ }
+ 
++#ifndef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
+ /**
```
ayazahma: why do we have these ndef macro

patches/0001-drm-pagemap-Disable-THP-migration-fall-back-to-page-.patch
```diff
@@ -0,0 +1,192 @@
...
+ 		/*
+ 		 * Some pages to migrate. But we want to migrate all or
+ 		 * nothing. Raced or unknown device pages.
+@@ -666,6 +681,7 @@
+ 		}
+ 		migrate.dst[i] = migrate_pfn(migrate.dst[i]);
+ 
++#ifndef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
+ 		if (migrate.src[i] & MIGRATE_PFN_COMPOUND) {
+ 			drm_WARN_ONCE(dpagemap->drm, src_page &&
```
ayazahma: same here

patches/0001-drm-pagemap-Disable-THP-migration-fall-back-to-page-.patch
```diff
@@ -0,0 +1,192 @@
...
+ 		else
+ 			folio = folio_alloc(GFP_HIGHUSER, order);
+ 
+@@ -801,8 +822,10 @@
+ 		page = folio_page(folio, 0);
+ 		mpfn[i] = migrate_pfn(page_to_pfn(page));
+ 
++#ifndef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
+ 		if (order)
+ 			mpfn[i] |= MIGRATE_PFN_COMPOUND;
```
ayazahma: same here

patches/0001-drm-pagemap-compat-for-older-kernels.patch
```diff
@@ -0,0 +1,192 @@
...
+@@ -791,7 +808,11 @@
+ 
+ 		/* TODO: Support fallback to single pages if THP allocation fails */
+ 		if (vas)
++#ifdef BPM_VMA_ALLOC_FOLIO_4ARGS_NOT_PRESENT
++			folio = vma_alloc_folio(GFP_HIGHUSER, order, vas, addr, false);
++#else
+ 			folio = vma_alloc_folio(GFP_HIGHUSER, order, vas, addr);
++#endif
+ 		else
```
ayazahma: should be managed in .h file

autotools/m4/drm_pagemap_folio_unlock.m4
```diff
@@ -0,0 +1,18 @@
+dnl #
+dnl # 3a5a06554566
+dnl # mm/zone_device: rename page_free callback to folio_free
+dnl #
```
ayazahma: pls dont do cleanup, I see soem issue in patches itself, need separate review

autotools/m4/list_lru_add.m4
```diff
@@ -0,0 +1,63 @@
...
+                                [Define to 1 if list_lru_add has 4 arguments])
+                ])
+        ])
+])
+
+dnl #
+dnl # fb56fdf8b9a2
+dnl # mm/list_lru: split the lock to per-cgroup scope
+dnl #
+AC_DEFUN([AC_LIST_LRU_ONE_LOCK_NOT_PRESENT], [
```
ayazahma: need to be split

ayazahma: based on commit

autotools/m4/list_lru_add.m4
```diff
@@ -0,0 +1,63 @@
...
+                ],[
+			struct list_lru lru;
+			struct list_head item;
+			int nid = 0;
+
+			list_lru_add(&lru, &item, nid, NULL);
+                ],[
+                        AC_DEFINE(BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT, 1,
+                                [Define to 1 if list_lru_add has 4 arguments])
+                ])
```
ayazahma, line 19: comment is misleading, not present means its less than 4

autotools/m4/list_lru_add.m4
```diff
@@ -0,0 +1,63 @@
...
+		AC_KERNEL_TRY_COMPILE([
+			#include <linux/list_lru.h>
+			#include <linux/spinlock.h>
+		],[
+			struct list_lru_one l;
+			spin_lock(&l.lock);
+		],[
+			AC_DEFINE(BPM_LIST_LRU_ONE_LOCK_NOT_PRESENT, 1,
+				[Define if struct list_lru_one has lock spinlock member])
+		])
```
ayazahma: same

ayazahma: Consistency in comment

autotools/m4/list_lru_add.m4
```diff
@@ -0,0 +1,63 @@
...
+						   void *arg) {
+				return LRU_REMOVED;
+			}
+		],[
+			list_lru_walk_cb fn = test_cb;
+			(void)fn;
+		],[
+			AC_DEFINE(BPM_HAVE_LIST_LRU_WALK_CB_NOT_PRESENT, 1,
+				[Define if list_lru walk callback has no spinlock parameter])
+		])
```
ayazahma: macro is emphasising that CB is not present

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+ * across distributions.
+ *
+ * BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT is defined by the backport build system via a
+ * compile-time probe when the 4-arg signature is detected. This correctly handles:
+ *   6.6 LTS    – 2-arg list_lru_add → call with 2 args
+ *   Ubuntu 6.8 – 4-arg list_lru_add → pass all 4 args through
+ *   6.12 LTS   – 4-arg list_lru_add → pass all 4 args through
+ *   7.0+       – 4-arg list_lru_add → pass all 4 args through
+ */
+
```
ayazahma: whats is this, I need information with commit mapping instead of kv

ayazahma: no kv ref , only commit ref

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+#undef list_lru_add
+
+#ifdef BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT
+#ifdef BPM_LIST_LRU_ONE_LOCK_NOT_PRESENT
+/*
+ * New struct layout (ubuntu-7.0, redhat-10.1+): spinlock and atomic_long_t
+ * nr_items moved into struct list_lru_one.  list_lru_add() is not exported;
+ * inline the null-memcg path using public struct fields.
+ */
```
ayazahma: same here, no OSV /kernel ref

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+	}
+	spin_unlock(&l->lock);
+	return ret;
+}
+#else
+/*
+ * Old struct layout (6.8–6.12 LTS, ubuntu-6.8, backporting headers): spinlock
+ * is in struct list_lru_node, nr_items is long.  list_lru_add() may not be
+ * exported; inline the null-memcg path using public struct fields.
+ */
```
ayazahma: Same

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+ *   Ubuntu 6.8 – 4-arg list_lru_add → pass all 4 args through
+ *   6.12 LTS   – 4-arg list_lru_add → pass all 4 args through
+ *   7.0+       – 4-arg list_lru_add → pass all 4 args through
+ */
+
+#undef list_lru_add
+
+#ifdef BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT
+#ifdef BPM_LIST_LRU_ONE_LOCK_NOT_PRESENT
+/*
```
ayazahma: Why nested ?

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+ * 6.6 LTS: 2-arg list_lru_add, exported.
+ */
+static __always_inline bool
+backport_list_lru_add(struct list_lru *lru, struct list_head *item,
+			int nid, struct mem_cgroup *memcg)
+{
+	return list_lru_add(lru, item);
+}
+#endif /* BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT */
+
```
ayazahma: does it mean, we will never use code as it its for any kernel version even in 7.0/7.1 , which looks wrong to me

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+#define list_lru_add(lru, item, nid, memcg) \
+	backport_list_lru_add((lru), (item), (nid), (memcg))
+
+/*
```
ayazahma: why this one also

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+ * Old (pre-6.13) type: (item, list, lock, cb_arg)  – 4 args
+ * New (6.13+)    type: (item, list, cb_arg)         – 3 args
+ *
+ * BPM_HAVE_LIST_LRU_WALK_CB_NOT_PRESENT is used for distro kernels (e.g.
+ * Ubuntu 7.0) that may not report LINUX_VERSION_CODE >= 6.13 but have
+ * already dropped the spinlock parameter.
+ */
+#if defined(BPM_HAVE_LIST_LRU_WALK_CB_NOT_PRESENT) || \
+    LINUX_VERSION_CODE >= KERNEL_VERSION(6, 13, 0)
+	/* New kernel: callback signature is (item, list, cb_arg) */
```
ayazahma: why we need kernel check , will fail on OSVs like 9.8/9.10

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,124 @@
...
+	#define LIST_LRU_WALK_CALLBACK(name) \
+		static enum lru_status name(struct list_head *item, \
+					    struct list_lru_one *list, \
+					    void *cb_arg)
+#else
+	/* Old kernel: wrap 4-arg callback to call 3-arg impl */
+	#define LIST_LRU_WALK_CALLBACK(name) \
+		static enum lru_status name##_impl(struct list_head *item, \
+						   struct list_lru_one *list, \
```
ayazahma: again , else look like its not tested on same kernel as part of dkms testing

autotools/m4/mm_zone_gpu_stats.m4
```diff
@@ -0,0 +1,18 @@
...
+	AC_KERNEL_DO_BACKGROUND([
+		AC_KERNEL_TRY_COMPILE([
+			#include <linux/mmzone.h>
+		],[
+			int x = NR_GPU_ACTIVE;
+			int y = NR_GPU_RECLAIM;
+		],[
+			AC_DEFINE(BPM_NR_GPU_STATS_PRESENT, 1,
+				[NR_GPU_ACTIVE and NR_GPU_RECLAIM are available])
+		])
```
ayazahma, line 17: what is impact of not having thsi ?
	NR_GPU_ACTIVE,	/* Pages assigned to GPU objects */
	NR_GPU_RECLAIM,	/* Pages in shrinkable GPU pools */

backport/backport-include/linux/workqueue.h
```diff
@@ -17,8 +17,16 @@
 #define disable_work_sync cancel_work_sync
 #endif
 
+#ifdef BPM_SYSTEM_DFL_WQ_NOT_PRESENT
+#define system_dfl_wq system_unbound_wq
+#endif
+
 #ifdef BPM_SYSTEM_PERCPU_WQ_NOT_PRESENT
 #define system_percpu_wq system_wq
 #endif
```
ayazahma, line 25: not  releated to thsi backport but BPM_SYSTEM_DFL_WQ_NOT_PRESENT BPM_SYSTEM_PERCPU_WQ_NOT_PRESENT are from same commit

ayazahma (changes requested):

backport/backport-include/linux/gpu_buddy.h
```diff
@@ -0,0 +1,43 @@
+/* SPDX-License-Identifier: GPL-2.0 */
```
ayazahma, line 1: single commit ref for single patch

ayazahma: OK to have in case of full file copy

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+Date: Mon, 07 Jul 2026 00:00:00 +0530
+Subject: drm/xe: Add BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT guards for single-pass compat
+
+Guard two-pass MMU interval notifier code with BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+in xe_tlb_inval, xe_vm_types, xe_vm, and xe_vm_madvise so that older kernels
+without the two-pass interface can still compile.  Add xe_vm_range_tilemask_tlb_inval()
+as the single-pass fallback.
+
+Signed-off-by: Pravalika Gurram <pravalika.gurram@intel.com>
+---
```
ayazahma: missing ref commit

ayazahma: why MACRO is here

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+
+--- a/drivers/gpu/drm/xe/xe_tlb_inval_types.h
++++ b/drivers/gpu/drm/xe/xe_tlb_inval_types.h
+@@ -9,7 +9,9 @@
+ #include <linux/workqueue.h>
+ #include <linux/dma-fence.h>
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ #include "xe_device_types.h"
++#endif
```
ayazahma: Why Xe device is not needed in case of missing notifier ?

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+ 
+ struct drm_suballoc;
+ struct xe_tlb_inval;
+@@ -134,6 +136,7 @@
+ 	ktime_t inval_time;
+ };
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ /**
```
ayazahma: why we are removing code if notifier is present , which is latest xe on latest KMD

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+@@ -45,10 +45,12 @@
+ 
+ bool xe_tlb_inval_idle(struct xe_tlb_inval *tlb_inval);
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ int xe_tlb_inval_range_tilemask_submit(struct xe_device *xe, u32 asid,
+ 				       u64 start, u64 end, u8 tile_mask,
+ 				       struct xe_tlb_inval_batch *batch);
+ 
+ void xe_tlb_inval_batch_wait(struct xe_tlb_inval_batch *batch);
```
ayazahma: why ?

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
++++ b/drivers/gpu/drm/xe/xe_tlb_inval.c
+@@ -487,6 +487,7 @@
+ 	return list_is_singular(&tlb_inval->pending_fences);
+ }
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ /**
+  * xe_tlb_inval_batch_wait() - Wait for all fences in a TLB invalidation batch
+  * @batch: Batch of TLB invalidation fences to wait on
+@@ -570,3 +571,4 @@
```
ayazahma: Same

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+ }
++#endif
+--- a/drivers/gpu/drm/xe/xe_vm_types.h
++++ b/drivers/gpu/drm/xe/xe_vm_types.h
+@@ -18,7 +18,9 @@
+ #include "xe_device_types.h"
+ #include "xe_pt_types.h"
+ #include "xe_range_fence.h"
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ #include "xe_tlb_inval_types.h"
```
ayazahma: same

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,295 @@
...
+  * xe_vm_invalidate_vma_submit - Submit a job to invalidate GPU mappings for
+  * VMA.
+@@ -4490,7 +4491,71 @@
+ 	WRITE_ONCE(vma->tile_invalidated, vma->tile_mask);
+ 	return ret;
+ }
++#endif
+ 
++#ifdef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
++/**
```
ayazahma: look like broken patch, will break DKMS against 7.1 or later

patches/0001-drm-xe-xe_svm-mmu-notifier-single-pass.patch
```diff
@@ -0,0 +1,50 @@
...
++#endif
+ 	struct xe_device *xe = vm->xe;
+ 	struct drm_gpusvm_range *r, *first;
+ 	struct xe_tile *tile;
+@@ -278,10 +282,15 @@
+ 
+ 	xe_device_wmb(xe);
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ 	err = xe_tlb_inval_range_tilemask_submit(xe, vm->usm.asid, adj_start, adj_end,
```
ayazahma: need justification for any further review

patches/0001-drm-xe-userptr-mmu-notifier-single-pass-compat.patch
```diff
@@ -0,0 +1,186 @@
...
+ 	 * write: vm->svm.gpusvm.notifier_lock in read mode and vm->resv held.
+--- a/drivers/gpu/drm/xe/xe_userptr.c
++++ b/drivers/gpu/drm/xe/xe_userptr.c
+@@ -8,9 +8,12 @@
+ 
+ #include <linux/mm.h>
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ #include "xe_tlb_inval.h"
++#endif
```
ayazahma: again same

autotools/m4/drm_buddy_contiguous.m4
```diff
@@ -0,0 +1,16 @@
...
+                        #include <drm/drm_buddy.h>
+                ],[
+                        unsigned long flags = DRM_BUDDY_CONTIGUOUS_ALLOCATION;
+                        (void)flags;
+                ],[
+                        AC_DEFINE(BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT, 1,
+                                [DRM_BUDDY_CONTIGUOUS_ALLOCATION is defined in drm_buddy.h])
+                ])
+        ])
+])
```
ayazahma: shoudl be part of drm buddy chnages

patches/0001-drm-xe-firmware-signature-failure-plus-vram-mgr-follow-up.patch
```diff
@@ -0,0 +1,57 @@
...
++#endif
++
+ 	err = gpu_buddy_alloc_blocks(mm, (u64)place->fpfn << PAGE_SHIFT,
+ 				     (u64)lpfn << PAGE_SHIFT, size,
+ 				     min_page_size, &vres->blocks, vres->flags);
+ 	if (err)
+ 		goto error_unlock;
++#ifndef BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT
++	if (place->flags & TTM_PL_FLAG_CONTIGUOUS) {
++		if (!drm_buddy_block_trim(mm, NULL, vres->base.size, &vres->blocks))
```
ayazahma, line 55: same

patches/0001-drm-xe-firmware-signature-failure-plus-vram-mgr-follow-up.patch
```diff
@@ -0,0 +1,57 @@
...
++++ b/drivers/gpu/drm/xe/xe_ttm_vram_mgr.c
+@@ -81,9 +81,10 @@
+ 
+ 	if (place->flags & TTM_PL_FLAG_TOPDOWN)
+ 		vres->flags |= GPU_BUDDY_TOPDOWN_ALLOCATION;
+-
++#ifdef BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT
+ 	if (place->flags & TTM_PL_FLAG_CONTIGUOUS)
+ 		vres->flags |= GPU_BUDDY_CONTIGUOUS_ALLOCATION;
++#endif
```
ayazahma, line 30: BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT shoudl be set to NOT present in m4 file

autotools/m4/dev_pagemap_ops_folio_free.m4
```diff
@@ -15,3 +17,21 @@ AC_DEFUN([AC_DEV_PAGEMAP_OPS_FOLIO_FREE_NOT_PRESENT], [
 		])
 	])
 ])
+
+dnl #
+dnl # v6.7+ - folio_zone_device_data() added to linux/memremap.h
+dnl #
+AC_DEFUN([AC_FOLIO_ZONE_DEVICE_DATA_NOT_PRESENT], [
+       AC_KERNEL_DO_BACKGROUND([
```
ayazahma, line 24: we should rename the m4 file for all folio handling or have separate file form DATA NOT present

autotools/m4/dev_pagemap_ops_folio_free.m4
```diff
@@ -1,6 +1,8 @@
 dnl #
-dnl # v7.0 - struct dev_pagemap_ops gained folio_free callback
+dnl # da0c02516c50
```
ayazahma: noise

patches/0001-drm-ttm-list_lru_walk_node-arg3-compat.patch
```diff
@@ -0,0 +1,39 @@
...
+@@ -388,9 +386,7 @@ static void ttm_pool_type_init(struct ttm_pool_type *pt, struct ttm_pool *pool,
+ 	spin_unlock(&shrinker_lock);
+ }
+ 
+-static enum lru_status pool_move_to_dispose_list(struct list_head *item,
+-						 struct list_lru_one *list,
+-						 void *cb_arg)
++LIST_LRU_WALK_CALLBACK(pool_move_to_dispose_list)
+ {
+ 	struct list_head *dispose = cb_arg;
```
ayazahma: in complete backport or mixed with something else

backport/backport-include/drm/drm_suballoc.h
```diff
@@ -0,0 +1,76 @@
...
+/*
+ * Backport compatibility shim for drm_suballoc_alloc() and
+ * drm_suballoc_insert() added in kernel 7.1.
+ *
+ * In 7.1 the single drm_suballoc_new() was split into two phases:
+ *   drm_suballoc_alloc(gfp)               - allocate struct (no sleeping)
+ *   drm_suballoc_insert(mgr, sa, ...)     - wait for a free slot and insert
+ *
+ * On 6.6 we emulate both using the existing drm_suballoc_new().
+ */
```
ayazahma: comment need cleanup

backport/backport-include/drm/drm_suballoc.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef __BACKPORT_DRM_SUBALLOC_H__
+#define __BACKPORT_DRM_SUBALLOC_H__
+
+#include_next <drm/drm_suballoc.h>
+
+#ifdef BPM_DRM_SUBALLOC_ALLOC_NOT_PRESENT
+
```
ayazahma, line 9: may need to clean commit msg or comment since both header file and missing function share same name

backport/backport-include/drm/drm_suballoc.h
```diff
@@ -0,0 +1,76 @@
...
+ * @align:      required alignment (0 = manager default).
+ *
+ * Backport: calls drm_suballoc_new() to do the blocking wait and slot
+ * assignment, then moves the result into @sa and frees the temporary
+ * allocation that drm_suballoc_new() created internally.
+ */
+static inline int drm_suballoc_insert(struct drm_suballoc_manager *sa_manager,
+				      struct drm_suballoc *sa, size_t size,
+				      bool intr, size_t align)
+{
```
ayazahma, line 59: why we are adding this function based on Missing "SUBALLOC_ALLOC"

backport/backport-include/drm/drm_suballoc.h
```diff
@@ -0,0 +1,76 @@
...
+	sa->manager = tmp->manager;
+	sa->soffset = tmp->soffset;
+	sa->eoffset = tmp->eoffset;
+	sa->fence   = tmp->fence;
+
+	kfree(tmp);
+	return 0;
+}
+
+#endif /* BPM_DRM_SUBALLOC_ALLOC_NOT_PRESENT */
```
ayazahma, line 82: in original patch , these new functions are exported as Symbol but missing in your implementation

backport/backport-include/linux/bitmap.h
```diff
@@ -3,21 +3,18 @@
 #define _BACKPORT_LINUX_BITMAP_H
 #include_next <linux/bitmap.h>
 
-#ifdef BPM_BITMAP_WEIGHTED_OR_NOT_PRESENT
-/* bitmap_weighted_or - return the weight of the OR of two bitmaps
```
ayazahma: why removed BPM_BITMAP_WEIGHTED_OR_NOT_PRESENT

ayazahma (changes requested):

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: #define XE_MAX_GT_PER_TILE 2 -t is defined in xe_device_types

For single pass communication below code is not defined so xe_device_types.h is not needed
int xe_vm_range_tilemask_tlb_inval(struct xe_vm *vm, u64 start,
+               u64 end, u8 tile_mask)
+{
+       struct xe_tlb_inval_fence
+               fence[XE_MAX_TILES_PER_DEVICE * XE_MAX_GT_PER_TILE];
+       struct xe_tile *tile;
+       u32 fence_id = 0;

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT will false 
#ifndef 0
// This code will be present in latest where notifier is present
#endif

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT will false in case of invalidate_start_and invalidate_finish present
#ifndef 0
// This code will be present in latest where notifier is present
#endif

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT will false in case of invalidate_start_and invalidate_finish present
#ifndef 0
// This code will be present in latest where notifier is present
#endif

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT will false in case of invalidate_start_and invalidate_finish present
#ifndef 0
// This code will be present in latest where notifier is present
#endif

patches/0001-drm-xe-mmu-notifier-single-pass-compat.patch
pravalikagurram1: BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT will false in case of invalidate_start_and invalidate_finish present
#ifndef 0
// This code will be present in latest where notifier is present
#endif
No it will not break 7.1 ,  in case of 7.1 this macro itself will not be enabled 
Then 
#ifdef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
// this code is for single pass communication so this will not enter into this so code will not be enabled
#endif

patches/0001-drm-xe-userptr-mmu-notifier-single-pass-compat.patch
pravalikagurram1: same above explanation applies here

patches/0001-drm-xe-xe_svm-mmu-notifier-single-pass.patch
pravalikagurram1: Below 7.1
BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT = 1
Above =7.1
BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT not defined

#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
// This code will be enabled in case of 7.1 because this above macro is no enabled since its negate condition 
In thsi section am using TWO_PASS communication
#endif

#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
//SINGLE PASS communication will only enabled when BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT enabled
>= 7.1 This will not be enabled
#nedif

patches/0001-drm-gpusvm-disable-thp-migration-on-non-folio-kernels.patch
pravalikagurram1: This we will do in Cleanup change . getting hunk what ever we change,
 would like to work on cleanup of drm_pagemap.c. after this merge will take care in that. I have created patch but getting hunk on other these are tightly coupled .

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
ayazahma: 4516432284e1b2ad9e70de8067f779c9c1072189 add the Change of nid and you are adopting that in your patch

ayazahma: ttm/pool: track allocated_pages per numa node.

backport/backport-include/linux/bitmap.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef _BACKPORT_LINUX_BITMAP_H
+#define _BACKPORT_LINUX_BITMAP_H
+#include_next <linux/bitmap.h>
+
+#ifdef BPM_BITMAP_WEIGHTED_OR_FUNC_NOT_PRESENT
+/* bitmap_weighted_or - return the weight of the OR of two bitmaps
+ * Similar to bitmap_or() but returns the weight (number of bits set)
+ * instead of just performing the operation.
+ * This function performs: dst = src1 | src2 and returns popcount(dst)
+ */
```
ayazahma: remove un nacessary comment

autotools/m4/bitmap_weighted.m4
```diff
@@ -0,0 +1,32 @@
...
```
ayazahma, line 1: for all commits lets stop using autotools/bacport

its may be 
compat/bitmap: xxx xxx xxx all commity sneed to be updated

backport/backport-include/linux/bitmap.h
```diff
@@ -0,0 +1,23 @@
...
+static inline unsigned int bitmap_weighted_or(unsigned long *dst,
+		const unsigned long *src1,
+		const unsigned long *src2,
+		unsigned int nbits)
+{
+	bitmap_or(dst, src1, src2, nbits);
+	return bitmap_weight(dst, nbits);
+}
```
ayazahma, line 22: does the final implementation are doing the same in both case

autotools/m4/dma_buf.m4
```diff
@@ -0,0 +1,54 @@
+dnl #
+dnl # 95308225e5ba
+dnl # dma-buf: Rename dma_buf_move_notify() to dma_buf_invalidate_mappings()
+dnl #
+
```
ayazahma, line 5: need sha for 2nd  cachage

autotools/m4/dma_fence.m4
```diff
@@ -0,0 +1,41 @@
...
+		])
+	])
+])
+
+dnl #
+dnl # 
+dnl #
+dnl #
+AC_DEFUN([AC_DMA_FENCE_LOCK_IRQSAVE_AND_IRQRESTORE_NOT_PRESENT], [
+	AC_KERNEL_DO_BACKGROUND([
```
ayazahma, line 27: need sha for thsi one also

autotools/m4/mmu_interval_notifier_twopass.m4
```diff
@@ -0,0 +1,21 @@
+dnl #
```
ayazahma, line 1: commit msg need  to be chnaged

patches/0001-drm-xe-fallback-implementation-for-single-pass-MMU-n.patch
```diff
@@ -0,0 +1,534 @@
...
++#endif
+ 	struct xe_device *xe = vm->xe;
+ 	struct drm_gpusvm_range *r, *first;
+ 	struct xe_tile *tile;
+@@ -280,10 +284,15 @@ static void xe_svm_invalidate(struct drm_gpusvm *gpusvm,
+ 
+ 	xe_device_wmb(xe);
+ 
++#ifndef BPM_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT
+ 	err = xe_tlb_inval_range_tilemask_submit(xe, vm->usm.asid, adj_start, adj_end,
```
ayazahma, line 69: why check #ifndef

autotools/m4/mm_zone_gpu_stats.m4
pravalikagurram1: This is what i got info
no functional impact but still we need to see in future if these node will be used by any UMD to track. This will be the limitation. Do i need to talk with experts here?

"use gpu mm stats to track gpu memory allocations. (v4)

    This uses the newly introduced per-node gpu tracking stats,
    to track GPU memory allocated via TTM and reclaimable memory in
    the TTM page pools.

    These stats will be useful later for system information and
    later when mem cgroups are integrated."

ayazahma (changes requested):

patches/0001-drm-pagemap-zone_device_page_init-1arg-on-older.patch
```diff
@@ -1,27 +0,0 @@
...
- drivers/gpu/drm/drm_pagemap.c | 4 ++++
- 1 file changed, 4 insertions(+)
-
-diff --git a/drivers/gpu/drm/drm_pagemap.c b/drivers/gpu/drm/drm_pagemap.c
---- a/drivers/gpu/drm/drm_pagemap.c
-+++ b/drivers/gpu/drm/drm_pagemap.c
-@@ -201,5 +201,9 @@ static void drm_pagemap_get_devmem_page(struct page *page,
- 				    struct drm_pagemap_zdd *zdd)
- {
- 	page->zone_device_data = drm_pagemap_zdd_get(zdd);
```
pravalikagurram1, line 19: This removal is needed because the below code inside the function is not present but its  needed for THP fall back option, hence adding it after that THP disablement

patches/0001-drm-pagemap-zone_device_page_init-1arg-on-older.patch
```diff
@@ -1,27 +0,0 @@
...
-+#ifdef BPM_ZONE_DEVICE_PAGE_INIT_3ARGS_NOT_PRESENT
-+	zone_device_page_init(page);
-+#else
- 	zone_device_page_init(page, page_pgmap(page), 0);
-+#endif
- }
```
ayazahma, line 25: based on code analaysis,  zone_device_page_init() is no longer called from xe or drm_pagemap , so we should remove correspondiong m4 files also

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
ayazahma, line 7: atches: Fix shrinker debugfs format to include NUMA node ID
Update the seq_printf format string in the shrinker compat patch to print the
NUMA node ID, matching the format used in the else path.

do I need to understand else path ? does this else path is part of base kernel or added by base kernel.

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
```diff
@@ -47,42 +50,42 @@ index baf27c7..8440f27 100644
...
+ 	for_each_node(nid) {
+ 		sc.nid = nid;
 +#ifdef BPM_SHRINKER_ALLOC_NOT_PRESENT
-+	count = ttm_pool_shrinker_count(&mm_shrinker, &sc);
-+	seq_printf(m, "%lu/%lu\n", count,
-+		   ttm_pool_shrinker_scan(&mm_shrinker, &sc));
++		count = ttm_pool_shrinker_count(&mm_shrinker, &sc);
++		seq_printf(m, "%lu/%lu\n", count,
++			   ttm_pool_shrinker_scan(&mm_shrinker, &sc));
 +#else
```
ayazahma, line 60: how this cleanup is related rebase, we should not mix rebase  and cleanup

autotools/m4/bitmap_weighted.m4
```diff
@@ -0,0 +1,32 @@
...
+dnl # 437cb3ded25
+dnl # cpumask: Introduce cpumask_weighted_or()
+dnl #
+AC_DEFUN([AC_BITMAP_WEIGHTED_OR_NOT_PRESENT], [
```
ayazahma, line 5: why single commit information and single patch

backport/backport-include/linux/dma-buf.h
```diff
@@ -21,4 +21,13 @@ MODULE_IMPORT_NS(DMA_BUF);
...
 
+#ifdef BPM_DMA_BUF_INVALIDATE_MAPPINGS_NOT_PRESENT
+/*
+ * dma_buf_invalidate_mappings - notify importers to invalidate their mappings
+ * @dma_buf: buffer whose mappings need to be invalidated
+ *
+ */
+static inline void dma_buf_invalidate_mappings(struct dma_buf *dma_buf) {}
+#endif /* BPM_DMA_BUF_INVALIDATE_MAPPINGS_NOT_PRESENT */
+
```
ayazahma, line 30: This is like providing dummy definition not mapping to any thing

autotools/m4/dma_buf.m4
```diff
@@ -0,0 +1,26 @@
+dnl #
```
ayazahma, line 1: map dma_buf_move_notify() to dma_buf_invalidate_mappings for mainline kernel x.y.z-  --> option one
Do we have still process fo listing compilation error ?

autotools/m4/dma_buf.m4
ayazahma, line 1: same needed for thsi commit

autotools/m4/dma_fence.m4
```diff
@@ -0,0 +1,20 @@
+dnl #
```
ayazahma, line 1: commitg msg

autotools/m4/dma_fence.m4
```diff
@@ -18,3 +18,24 @@ AC_DEFUN([AC_DMA_FENCE_EXTERN_LOCK_NOT_PRESENT], [
 		])
 	])
 ])
+
+dnl #
```
ayazahma, line 22: why ref is missing here

ayazahma: in case if its part of some other sha handling , pls there with single macro

autotools/m4/dma_fence.m4
```diff
@@ -18,3 +18,24 @@ AC_DEFUN([AC_DMA_FENCE_EXTERN_LOCK_NOT_PRESENT], [
...
+AC_DEFUN([AC_DMA_FENCE_LOCK_IRQSAVE_AND_IRQRESTORE_NOT_PRESENT], [
+	AC_KERNEL_DO_BACKGROUND([
+		AC_KERNEL_TRY_COMPILE([
+			#include <linux/dma-fence.h>
+		],[
+			struct dma_fence *fence = NULL;
+			unsigned long flags;
+			dma_fence_lock_irqsave(fence, flags);
+			dma_fence_unlock_irqrestore(fence, flags);
+		],[
```
ayazahma, line 35: if we want a handler for each missing function then we should have a macro for each check

autotools/m4/dma_fence.m4
ayazahma: 1f32f310a13 has not introduced this change so ref need to be corrected
need to understand in details as there seems to be risk of partial locking.

autotools/m4/list_lru_add.m4
```diff
@@ -0,0 +1,21 @@
+dnl #
```
ayazahma, line 1: Add list_lru_add function use 2args instead 4
handle list_lru_add () arg mismatch or alignments   

commit msg quality

autotools/m4/list_lru_add_obj.m4
```diff
@@ -0,0 +1,19 @@
+dnl #
```
ayazahma, line 1: coomit msg shoudl not talk about obious code change by why , and what wil lbe impact etc with analysis

backport/backport-include/linux/list_lru.h
```diff
@@ -12,4 +12,25 @@
 #define list_lru_add(lru, item, nid, memcg) list_lru_add_obj(lru, item)
 #endif
 
+#ifdef BPM_HAVE_LIST_LRU_WALK_CB_NO_SPINLOCK
+	#define LIST_LRU_WALK_CALLBACK(name) \
+		static enum lru_status name(struct list_head *item, \
+					    struct list_lru_one *list, \
+					    void *cb_arg)
+#else
```
ayazahma: where the is the case of 7.1 or OSV where da0c02516c50  is backported by OSV /LTS ?

patches/0001-drm-ttm-guard-nr_gpu-accounting.patch
```diff
@@ -0,0 +1,68 @@
...
+--- a/drivers/gpu/drm/ttm/ttm_pool.c
++++ b/drivers/gpu/drm/ttm/ttm_pool.c
+@@ -185,7 +185,9 @@ static struct page *ttm_pool_alloc_page(struct ttm_pool *pool, gfp_t gfp_flags,
+ 		p = alloc_pages_node(pool->nid, gfp_flags, order);
+ 		if (p) {
+ 			p->private = order;
++#ifdef BPM_NR_GPU_STATS_PRESENT
+ 			mod_lruvec_page_state(p, NR_GPU_ACTIVE, 1 << order);
++#endif
+ 		}
```
ayazahma, line 30: now on older kernel, we are not calling "mod_lruvec_page_state" , was the functional correctness or impact analysis is done ?

autotools/m4/system_dfl_wq.m4
```diff
@@ -0,0 +1,25 @@
+dnl #
```
ayazahma: commit msgs shoudl have explanation for impact if needed to explain the change

autotools/m4/xe.m4
```diff
@@ -98,6 +98,7 @@ dnl # SVM-related probes
 	AC_VMA_ALLOC_FOLIO_4ARGS_NOT_PRESENT
 	AC_ZONE_DEVICE_PAGE_INIT_3ARGS_NOT_PRESENT
 	AC_MIGRATE_DEVICE_PFNS_NOT_PRESENT
+	AC_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT
```
ayazahma, line 107: BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_PRESENT --> BPM_DRM_BUDDY_CONTIGUOUS_ALLOCATION_NOT_PRESENT

backport/backport-include/linux/memremap.h
```diff
@@ -0,0 +1,22 @@
...
+ * @folio: the folio
+ *
+ * In order to access zone_device_data we can utilize
+ * folio->page.zone_device_data directly.
+ */
+static inline void *folio_zone_device_data(const struct folio *folio)
+{
+	return folio->page.zone_device_data;
+}
+#endif /* BPM_FOLIO_ZONE_DEVICE_DATA_NOT_PRESENT */
```
ayazahma, line 14: static inline void *folio_zone_device_data(const struct folio *folio)
{
	VM_WARN_ON_FOLIO(!folio_is_device_private(folio), folio);
	return folio->page.zone_device_data;
}|

why Warn On is not needed ?

autotools/m4/mmu_interval_notifier_twopass.m4
```diff
@@ -0,0 +1,21 @@
...
+dnl # 7aba71dbc416
+dnl # mm/mmu_notifier: Allow two-pass struct mmu_interval_notifiers
+dnl #
+
+AC_DEFUN([AC_MMU_INTERVAL_NOTIFIER_TWOPASS_NOT_PRESENT], [
+        AC_KERNEL_DO_BACKGROUND([
+                AC_KERNEL_TRY_COMPILE([
+                    #include <linux/mmu_notifier.h>
+                ],[
```
ayazahma, line 10: need further review , will revert back
Pravalika to provide impact analysis

patches/0001-drm-pagemap-Disable-THP-migration-fall-back-to-page-.patch
```diff
@@ -0,0 +1,176 @@
...
+ }
+ 
+ /**
+@@ -455,6 +460,7 @@ out:
+ 	return ret;
+ }
+ 
++#ifndef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
+ /**
+  * drm_pagemap_cpages() - Count collected pages
```
ayazahma: why a code is commented for not def , need explanation

patches/0001-drm-pagemap-Disable-THP-migration-fall-back-to-page-.patch
```diff
@@ -0,0 +1,176 @@
...
+ 		.vma		= vas,
+ 		.pgmap_owner	= page_pgmap(page)->owner,
++#ifdef BPM_ZONE_DEVICE_FOLIO_INIT_NOT_PRESENT
++		.flags		= MIGRATE_VMA_SELECT_DEVICE_PRIVATE |
++				  MIGRATE_VMA_SELECT_DEVICE_COHERENT, 
++#else
+ 		.flags		= MIGRATE_VMA_SELECT_DEVICE_PRIVATE |
+ 				  MIGRATE_VMA_SELECT_DEVICE_COHERENT |
+ 				  MIGRATE_VMA_SELECT_COMPOUND,
++#endif
```
ayazahma: what will be impact of not having COMPOUND VMA?

autotools/m4/bitmap_weighted_or.m4
```diff
@@ -0,0 +1,32 @@
...
+			AC_KERNEL_TRY_COMPILE_SYMBOL([
+			#include <linux/bitmap.h>
+			],[
+			unsigned long dst, src1 = 1, src2 = 2;
+			unsigned int weight = __bitmap_weighted_or(&dst, &src1, &src2, 32);
+			(void)weight;
+			],[__bitmap_weighted_or],[lib/bitmap.c],[
+			],[
+			AC_DEFINE([BPM_BITMAP_WEIGHTED_OR_NOT_PRESENT], 1,
+				[__bitmap_weighted_or() is not exported by the kernel])
```
ayazahma: one is __Bitmap -xxxx other is bitmap_xxxx , but macro is showing them as function and variable

ayazahma (changes requested):

autotools/m4/dma_fence.m4
ayazahma: or we can have single macro for all the changes of a commit

autotools/m4/dma_buf.m4
akanksh5-intc: Understanding:

1. If we have 2 different SHA we need two different commit messages with two different usages of macro.
2.  And if the changes are already part of a single commit in upstream, then we can handle using single macro.

ayazahma (changes requested):

backport/backport-include/linux/dma-buf.h
```diff
@@ -33,4 +33,8 @@ static inline void dma_buf_invalidate_mappings(struct dma_buf *dma_buf)
 }
 #endif /* BPM_DMA_BUF_INVALIDATE_MAPPINGS_NOT_PRESENT */
 
+#ifdef BPM_DMA_BUF_ATTACH_OPS_INVALIDATE_MAPPINGS_NOT_PRESENT
+#define invalidate_mappings move_notify
```
smuqthya, line 32: I would like to have test coverage is done or not for this. As the caller will check if the fn ptr is assigned or not and return back if not assigned. At compile time we may see the assignment proper, but at run time, in-built dma-buf may call move_notify and assumes its not assigned and go back. So its better to have validated.

smuqthya: Another way to verify. Generate a preprocessor file (.i) for this .c  and see if replacement at caller level is proper.

backport/backport-include/linux/dma-fence.h
```diff
@@ -19,4 +19,8 @@ dma_fence_check_and_signal_locked(struct dma_fence *fence)
 }
 #endif /* BPM_DMA_FENCE_CHECK_AND_SIGNAL_LOCKED_NOT_PRESENT */
 
+#ifdef BPM_DMA_FENCE_EXTERN_LOCK_NOT_PRESENT
+#define extern_lock lock
```
smuqthya, line 23: Lets verify the coverage or preprocessor way to confirm. and how are we sure this is not going out side dma_fence structure variable initialization or usage. and what is the impact of replacement outside dma_fence

smuqthya: Also we need to check impact of inline lock vs external lock and how replacing all with external will be enough to cover full functionality? May be as a post merge task

backport/backport-include/linux/list_lru.h
```diff
@@ -0,0 +1,13 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+
+#ifndef __BACKPORT_LIST_LRU_H__
+#define __BACKPORT_LIST_LRU_H__
+
+#include <linux/version.h>
+#include_next <linux/list_lru.h>
+
+#ifdef BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT
+#define list_lru_add(lru, item, nid, memcg) list_lru_add(lru, item)
```
smuqthya, line 10: impact of not handling -  nid, memcg

smuqthya: Should be using - list_lru_add_obj

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
akanksh5-intc: Done

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
akanksh5-intc: Not mentioning now

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
akanksh5-intc: Noted and resolved

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
akanksh5-intc: Resolved and adopting in other patches too.

patches/0001-autotools-m4-manage-shrinker-API-changes-from-old-to.patch
akanksh5-intc: Done

patches/0002-Add-debug-print-and-BACKPORT_MOD_VERSIONS-for-oot-mo.patch
akanksh5-intc: Addressed.

autotools/m4/bitmap_weighted.m4
akanksh5-intc: Done

backport/backport-include/linux/list_lru.h
```diff
@@ -8,6 +8,8 @@
 
 #ifdef BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT
 #define list_lru_add(lru, item, nid, memcg) list_lru_add(lru, item)
-#endif /* BPM_LIST_LRU_ADD_4ARGS_NOT_PRESENT */
+#elif defined(BPM_LIST_LRU_ADD_OBJ_NOT_PRESENT)
+#define list_lru_add(lru, item, nid, memcg) list_lru_add_obj(lru, item)
```
smuqthya, line 12: This is not correct. should be opposite order

No need to of separate M4. lets use previous one. 

#define list_lru_add_obj(lru, item) list_lru_add(lru, item)

smuqthya (approved):

smuqthya (changes requested):
with updated comments and 2 changes

smuqthya (approved):

PR #487, opened 2026-08-05
drm/xe/eudebug: Avoid taking discovery lock during teardown

kbommu:
Display teardown closes the DRM client, causing eudebug to send a
client-close event. Previously, this required taking the discovery
lock, creating a circular locking dependency.

Avoid taking the discovery lock in file_close() and rely on
xe->clients.lock, which already protects the client list.

Since the client list may change during discovery, create a temporary
list with client references and iterate over that instead.

This removes the circular dependency while keeping client discovery
safe.


smuqthya (approved):

smuqthya:
I have seen the premerge, its not populating. but cross check before merge

PR #511, opened 2026-08-20
TODO: disable xe_i2c compilation based on designware_i2c header

pravalikagurram1:
The xe_i2c will be enabled once we have designware support in kernel,
which is not available in all kernel versions. Conditionally disable xe_i2c
compilation at the Makefile level using a wildcard check for the
header, and guard the function declarations in xe_i2c.h.

This will be enabled once full i2c backport support is added.


patches/0001-disable-xe_i2c-compilation-based-on-designware_i2c-h.patch
```diff
@@ -0,0 +1,57 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Pravalika Gurram <pravalika.gurram@intel.com>
+Date: Thu, 20 Aug 2026 17:24:52 +0530
+Subject: disable xe_i2c compilation based on designware_i2c header
```
smuqthya: Add TODO

smuqthya (approved):

PR #514, opened 2026-08-21
drm/xe: dma-buf Add VRAM pin limit for backports when dmem_cgroup is not present

pravalikagurram1:
Remove the dummy drmm_cgroup_register_region() stub from the backport
header and replace it with a driver-level VRAM pin accounting and
enforcement mechanism for kernels where dmem_cgroup is not available.

The dmem_cgroup subsystem, introduced in
newer kernels, provides per-cgroup limits on device memory usage such
as GPU VRAM. On older backport target kernels this subsystem is absent
(BPM_DRMM_CGROUP_REGISTER_REGION_NOT_PRESENT), so no VRAM pinning
limits were enforced — external dma-buf consumers could pin an
unbounded amount of VRAM, potentially starving the GPU driver and
local workloads.

The VRAM pin percentage is configurable via the module parameter:
  xe.pin_vram_percent=<0-100>  (default: 50)
A value of 0 disables the limit (unlimited pinning).

### DIFF
--- a/patches/0001-drm-xe-vram-show-pinned-vram-in-debug.patch
+++ b/patches/0001-drm-xe-vram-show-pinned-vram-in-debug.patch
@@ -1,0 +1,58 @@
+From 9adf8d7f4ca55cffdbe2cc58a023d2dae6c98cb0 Mon Sep 17 00:00:00 2001
+From: Himal Prasad Ghimiray <himal.prasad.ghimiray@intel.com>
+Date: Mon, 3 Aug 2026 10:40:35 +0530
+Subject: drm/xe/vram: show pinned vram in debug
+
+Expose the pinned VRAM counters and the dma-buf limit in the VRAM
+manager debug output.
+
+Reviewed-by: Tejas Upadhyay tejas.upadhyay@intel.com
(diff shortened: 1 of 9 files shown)


patches/0001-drm-xe-vram-show-pinned-vram-in-debug.patch
```diff
@@ -0,0 +1,58 @@
+From 9adf8d7f4ca55cffdbe2cc58a023d2dae6c98cb0 Mon Sep 17 00:00:00 2001
+From: Himal Prasad Ghimiray <himal.prasad.ghimiray@intel.com>
+Date: Mon, 3 Aug 2026 10:40:35 +0530
+Subject: [PATCH 7/8] drm/xe/vram: show pinned vram in debug
```
smuqthya: remove [Patch]

smuqthya (approved):

patches/0001-drm-xe-vram-show-pinned-vram-in-debug.patch
smuqthya: What about validation.

pravalikagurram1: validation kunit is not supported . Till igt comes

PR #538, opened 2026-09-01
xekmd-backport: fix hunk failure and compilation issue against supported osv's

akanksh5-intc:


backport/backport-include/linux/gpu_buddy.h
```diff
@@ -31,6 +31,7 @@
 #define gpu_buddy_print_tree drm_buddy_print_tree
 #define gpu_buddy_block_trim drm_buddy_block_trim
 #define gpu_buddy_driver_set_lock(mm, lock) do { (void)(mm); (void)(lock); } while (0)
+#define gpu_buddy_allocated_addr_to_block(mm, addr) ({ (void)(mm); (void)(addr); NULL; })
```
knayana-gse: instead of dummy op, can't we backport this api in out compat layer. any difficulties we are facing here? what is the impact if we are going with this dummy-op?

knayana-gse (changes requested):

backport/backport-include/linux/gpu_buddy.h
knayana-gse: can we replace with any alternative api without missing the functionality.

akanksh5-intc: I checked, there is no alternative api. Actually, this function was introduced from this patch "**backport/patches/base/0001-drm-gpu-Add-gpu_buddy_allocated_addr_to_block-helper.patch**", from recent merged pr in kernel-backport/main branch.

akanksh5-intc: I checked the feasibility, yes we can backport that api in compat layer under gpu_buddy.c file. Will give a try and post if some challenges faced.

knayana-gse (approved):
LGTM. Please check against supported OSVs.

akanksh5-intc:
> LGTM. Please check against supported OSVs.

Yes checked and compilation is clean.

PR #580, opened 2026-09-16
Fix __assign_str arg issue in xe_trace_cper.h

jagadeeshrb1:
__assign_str() accepts two arguments for below kernel 6.10.
Hence observing compilation issues with below kernel 6.10
To fix compilation issues, under macro condition, enabling 2 args for
older kernels.

Signed-off-by: Rachapalli Bandi Jagadeesh <jagadeesh.rachapalli.bandi@intel.com>


smuqthya (approved):
Cross verify once on all osv's

pravalikagurram1:
> Cross verify once on all osv's

Verified for all osv's

PR #585, opened 2026-09-18
[TEMP] Restore VF LMEM BAR size after disabling VFs

pravalikagurram1:
Backporting these changes, restores the PF VF_LMEM_BAR size during VF disable,
allowing VFs to be enabled again after an unplug/rescan cycle.

backported from https://patchwork.freedesktop.org/series/154724/#rev3


backport/patches/base/0001-drm-xe-pf-Keep-VF-LMEM-BAR-size-low-if-no-VFs-enable.patch
```diff
@@ -0,0 +1,74 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Marcin Bernatowicz <marcin.bernatowicz@linux.intel.com>
+Date: Wed, 2 Sep 2026 14:47:51 +0200
+Subject: drm/xe/pf: Keep VF LMEM BAR size low if no VFs enabled
```
ayazahma: I need [TEMP] in patch subject noty in pr as I'll never see pr in kernel patches 
add a line in commit to be reverted once merged in <target>

ayazahma (changes requested):

backport/patches/base/0001-drm-xe-pf-Keep-VF-LMEM-BAR-size-low-if-no-VFs-enable.patch
pravalikagurram1: Done

ayazahma (approved):

